PR #6, opened 2018-07-16
Add and fix compiler warnings.

MFraters:


MFraters:
For documentation: The coverage fell because the private method get_relative_path() lost coverage. This was because it was called without reason, and is not called anymore to fix the new compiler warnings. It can't (and probably shouldn't, see for example issue 258 of the catchorg/Catch2 repository on github) get unit tests because it is a private function. Because the function is not used, it should technically be removed, but I will keep it around for now.

PR #14, opened 2018-07-24
Add vtu program

MFraters:
Also added indenting to the app.


MFraters:
I didn't add tests for the the new visualization program or added it to the coverage testing because it is not part of the main library. It is currently not worth to invest time into testing it because it is only visualization. Might add testing later.

PR #26, opened 2018-07-27
Add linear temperature submodule to continental plate

MFraters:
This contains a bit more then advertised in the title. 

The most important non-oceanic plate change is that the point class doesn't have a default coordinate system anymore. Although it was very convenient in the beginning, it proved to be very error-prone and hard to debug when working with two coordinate systems. 

Secondly, the visualizer now has the option to output all the cells into the vtk file in 3d (for Cartesian and chunk). This make the vtk file a lot bigger (for 8 cells the difference is 27 vs 64), but it allows for better visualization in paraview. 

Furthermore some smaller changes, which I could have pulled out of this pull request, but in the interest of time and because they are all related to the development/debugging of proper functioning of the oceanic plate feature I kept it in one.


MFraters:
ah, I now see that this branch was about only adding a linear temperature submodule to the continental plate, so an other major difference is adding the oceanic plate feature.

PR #29, opened 2018-09-06
Add distance point from curved plane and subduction zones

MFraters:
This is a large change to the code to enable making subducting plates in both Cartesian and spherical coordinates. Is seems to work as described, although there are still some todo points for this pull request.

1. add tests for the subducting plate plugin
2. add code for 'angle_at_begin_segment_with_surface' option and add the associated tests, or decide to add that later.


MFraters:
1. Although there are still a lot of tester/coverage improvements to do, they can wait for now because the coverage of the important parts of the code is good (especially the new utilities function).
2.  Decided to forbid 'angle_at_begin_segment_with_surface' for now, so other things can move on quicker.
This problem has been moved to issue #30.

PR #34, opened 2018-09-12
Add more documetation and add copyright notices to files which still missed them.

MFraters:


MFraters:
Also let doxygen output the private variables and did some minor cleanup.

PR #37, opened 2018-09-15
Fixing the fortran example and the readme describing how to use it.

MFraters:


MFraters:
This was partly an overlooked part of the code from #32, and partly it make the readme description more general.

PR #43, opened 2018-09-18
Add an initial manual about how to use the GWB.

MFraters:


MFraters:
This should resolve issue #16

PR #48, opened 2018-09-21
Turn the Feature plugin system into a autoregistering factory system.

MFraters:


MFraters:
Set the tester to gcc 4.9, and added a cmake flag `-std=c++11`, to make sure the new code is compatible with compilers which only support cpp11. Found that the app and visulazation program used features introduced after cpp11.

MFraters:
Apparently the gcov verion matters for the coverage percentage. For consistency, turned it back to version 6 (along with gcc etc.). The new flag `-std=c++11` should be sufficient to ensure cpp11 compatibility.

MFraters:
Looked at the 'loss' of coverage, can't find it. It is probably because there are less relavant lines (1811 to 1808).

PR #67, opened 2018-10-15
add travis indenting check.

MFraters:


MFraters:
coveralls is down, but it is no problem for this pull request, because in the code only indenting (through the indenting script) has been changed.

PR #71, opened 2018-10-16
Fix adiabatic tempeature

MFraters:


MFraters:
Added codecov because coveralls has been down for a while now. Also added codacy, which provide code quality checks. Furthermore tired to add Coverity scan again, but it still doesn't work.

PR #96, opened 2019-01-21
fix cmake version file for aspect.

MFraters:
Need to use WORLD_BUILDER_SOURCE_DIR for external project.


MFraters:
fix for to #96

PR #100, opened 2019-01-22
add add_double_subduction_cookbook for the ELEFANT example

cedrict:


MFraters:
Thanks for adding this!

PR #109, opened 2019-02-14
update readme.md

MFraters:


MFraters:
Codacy seems to be stuck, but these are only documentation changes to the readme file which I checked through github edit. So merging it should be fine.

Issue #118, opened 2019-04-16
warnings when used inside ASPECT


tjhei:
I am getting hundreds of warnings of the style
```
Building CXX object CMakeFiles/aspect.dir/contrib/WorldBuilder/source/features/fault.cc.o
In file included from ../contrib/WorldBuilder/include/world_builder/features/fault.h:23:0,
                 from ../contrib/WorldBuilder/source/features/fault.cc:20:
../contrib/WorldBuilder/include/world_builder/features/interface.h:227:14: warning: ‘WorldBuilder::Features::never_actually_usedFault’ defined but not used [-Wunused-variable]
   static int never_actually_used##klass = make_sure_compilation_unit_referenced##klass();
              ^
../contrib/WorldBuilder/include/world_builder/features/fault.h:39:5: note: in expansion of macro ‘WB_REGISTER_FEATURE_HEADER’
     WB_REGISTER_FEATURE_HEADER(Fault)
     ^
```
This is using gcc 5.4.

MFraters:
I will look into it. What I remember is that I needed this system to have a plugin system in a library (with a application like ASPECT, this is not needed), but I never got those warnings.

MFraters:
I just rebuild aspect plus the world builder with gcc 8.2.0 in debug mode (now building release mode) and I see no warnings. Before I go and try it with an other gcc, are there any special flags you are using?

tjhei:
Nothing special. This is in debug mode and with deal.II 8.5 (see https://jenkins.tjhei.info/blue/organizations/jenkins/aspect-test/detail/aspect-test/117/pipeline#step-24-log-290 for details) and deal.II 9.0.1 (see https://jenkins.tjhei.info/blue/organizations/jenkins/aspect-test/detail/aspect-test/120/pipeline/23 )

tjhei:
If you need to reproduce go into a local ASPECT directory and run:
```
docker run --rm -it -v $PWD:/src tjhei/dealii:v9.0.1-full-v9.0.1-r5-gcc5
mkdir build; cd build; cmake -G Ninja /src && ninja
```

tjhei:
is this fixed now, @MFraters ?

MFraters:
from my point of view it is. So let's close this one.

PR #120, opened 2019-04-17
Remove static header variable

MFraters:
The solution for having a plugin system has been reported to cause problems with gcc 5 (#118 and see geodynamics/aspect#2613). Doing some more research on a different solution, I am now testing whether a linker flag (-Wl,--whole-archive) might also solve the problem of non-referenced compiler units causing the plugin system to fail. This solution might even better, because it will fully automatize the plugin system.

(re-pull request from #119, had incorrectly setup remote in new clone)


tjhei:
What are you trying to do with these variables? The ``whole-archive`` flag is mostly used for static libraries, how do you think this will fix the issue?

MFraters:
The fundemental problem I am trying to solve is explained here: https://stackoverflow.com/questions/5202142/static-variable-initialization-over-a-library. For aspect there is no problem, because it compiles  the code directly. But the world builder is also (actually mostly) designed to be used as a library. Because I used the solution in the last post on that page, I got an unused variable. 

Doing a bit more digging, I found that there is also a linker variable which forces the linking of the whole library (and thereby the automatic registration). This solution seems to work on linux, osx and windows on travis, but in appveyor it gives a linking error static that there are multiple definitions of std library functions like vector and new. I tried to see whether making the windows environment on appveyor more like travis solves the problem, but it doesn't. So I am for now also adding a windows release test on travis (used to be only debug).

tjhei:
> The fundemental problem I am trying to solve is explained here:

Is the program not referencing anything in the worldbuilder library or what is the issue?

Maybe a simpler solution is to have a ``WorldBuilder_Initialize()`` or something like this that the user needs to call?

MFraters:
This is what I understand the problem to be. To auto register all the plugins they need to be added to some kind of static variable. In a normal program it initializes all the static variables before the program start, thereby giving the program the opportunity to auto register the plugins. For a library, static variables are not automatically initialized/included, only when they are implicitly or explicitly referenced in every compilation unit. So to auto register plugins, I need to make sure every plugin is actually 'used', in order for it to be included. That is what the code did and gave the warning, because the 'use' of the plugin may not actually do anything. 

The approach I am now are going for is telling the compiler that it actually has to include the whole library, and that way it will initialize the static variables and auto register the plugins. This method is actually better since with the previous method, the registration was not fully automatic and the code is cleaner now. 

Also I seem to have fixed the problem in appveyor :)

MFraters:
@tjhei Do you have an alternative solution?
Since the problems with the testers seem to be fixed, I will start removing all the unused static variables and make a pull request to update the world builder code in aspect.

tjhei:
> Do you have an alternative solution?

I do not, sorry. If this works for you, sure, go ahead. 

Long term you might need to think about something that is less complicated for the user?

MFraters:
So for aspect the `whole-archive` is not needed, because it is compiled directly. So there should be no change for aspect. For the codes which use the library, they need to add two more linking flags. If they want to support osx, then they need will need a seperate case for that and add one flag (different) instead of two. I will document that, so it shouldn't be too hard not I have figured out how to do it.

Although I also want a more user friendly approach, from reading about this problem, I have the feeling that it is quite fundamental to how cpp libraries work at a minimum (it may differ per compiler what is done exactly I think). The method I used before seemed to work good enough, and I am still a bit surprised that it gave those warnings for gcc 5, because with gcc 8 and the intel compiler I didn't see those warnings, as well as with gcc 4.9 and gcc 6 which is used by the world builder tester, which also tests linking the world builder to an application. Since the new solution should solve the problem for gcc 5, I a fine with changing to this solution

tjhei:
> I am still a bit surprised that it gave those warnings for gcc 5, because with gcc 8 and the intel compiler I didn't see those warnings

yeah, problems like this are annoying. :-( Thanks for working on a fix!

MFraters:
@tjhei I have tested this version of the world builder with the docker instructions you gave and I don't see the warnings anymore. 

I have updated (and retested) the documentation as well, and it should be ready to merge when the testers pass.

PR #125, opened 2019-06-12
Fix subduction plate model

MFraters:
For the rebuttal of the paper looked again at the McKenzie 1970 paper which is used to implement the subducting plate plate model and found a bug in my implementation and what is likely an error in the paper itself. This pull request fixes the bug, cleans up the code and circumvents the possible error in the paper by using the depth we have directly instead of computing it with angles. This is also fundamentally better for this code because we do no have a single straight slab like in the paper, but potentially a segmented curved slab.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,7 +126,7 @@ matrix:
       osx_image: xcode10
       env: 
          - MAKE_FILE_GENERATOR="Unix Makefiles"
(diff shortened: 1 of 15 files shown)


MFraters:
Note that the total effect of these changes seems to range on the order of less than a degree to a few tens of degrees.

PR #134, opened 2019-09-16
link manual GWB version to the global repository version.

MFraters:
This makes sure that the version as shown in the manual is always consistent with the actual version of the code as represented in the VERSION file in the root of the repository. This does create a relative link, which means that the manual should always be build from a certain location. This therefore creates a restriction, but I am not sure if it will be a problem in practice.


MFraters:
looking at [wikibooks](https://en.wikibooks.org/wiki/LaTeX/Modular_Documents), it states that 

> When using relative paths within a LaTeX file imported by \input or \include, it is important to note that the paths are relative to the directory in which the main .tex file resides, not to the directory in which the included (or input) file is found.

Testing to compile the latex out the directory shows that it is not currently possible anyway because of the requirements of of `def.tex` and the images. So not sure whether the above is correct, but it is not getting any worse with this change.

Issue #138, opened 2019-11-22
Issues on install WorldBuilder on MacOS


lmx08:
Hi,
I'm now trying to install WorldBuilder on my Macbook, following the steps described in manual 2.2.1.2 and 2.2.2. Unfortunately, 2 errors occurred:
.........
[ 90%] Built target WorldBuilder
[ 95%] Linking CXX executable bin/WorldBuilderVisualization
[ 95%] Built target WorldBuilderApp
ld: library not found for -lomp
clang-9: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/WorldBuilderVisualization] Error 1
make[1]: *** [CMakeFiles/WorldBuilderVisualization.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 98%] Built target unit_test_world_builder
make: *** [all] Error 2
.......
Can you help with the problems? many thanks!

Cheers,
mgxliu

MFraters:
Hey mgxliu,

Thank you for raising an issue. Cmake says it cannot find the openmp library. 

The openmp library is needed for building the world builder visualizer, but not for the main library and app. So depending on what you want to do openmp might not be needed. Both the world build library and the world builder app have been successfully compiled and can already be used. 

Nonetheless, the instructions do not work in your case, so to help you getting also the visualizer working I will need some information on your system:

What version of OSX and Xcode are you using (you can easily find your xcode version by looking in the `/Aplications/` directory for a app called `Xcode-10.0.app` if you xcode is 10.0)?

Also, could you post the result the first time you ran `cmake ..` (if you still have this) and of the command `which clang` and `xcodebuild -find clang`.

Thanks,

menno

MFraters:
Looking a bit more into the problem, I remembered that at some point OSX changed their location of where clang is placed. Older versions used `usr/local/opt/llvm/` (which the manual assumes) and newer version use `/Applications/Xcode-XX.X.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/`. I am not sure between what version this changed happened exactly, but exporting the following flags should solve your problem:

`export LDFLAGS="-L/Applications/Xcode-XX.X.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib"`
`export CPPFLAGS="-I/Applications/Xcode-XX.X.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include"`
[...]

PR #139, opened 2019-11-22
Add newer osx tester

MFraters:
testing solutions to the problem described in #138.


MFraters:
I do not think this is needed for now.

PR #157, opened 2020-06-05
Enable and adress more warnings

MFraters:
It started out as addressing issue https://github.com/geodynamics/aspect/issues/3479, but I also enabled more warnings and fixed them (or am in the process of adressing them).


MFraters:
Somehow the coverage is not being posted (might take a while?), but it has not effectively changed (only dropped in one file because there are less lines), see: https://codecov.io/gh/GeodynamicWorldBuilder/WorldBuilder/pull/157/diff

PR #158, opened 2020-06-05
Improve temperature operation structure

MFraters:
This improves the temperature operation structure. It prevents having to check strings in every call, and instead just do a switch statments with enums.

Since this should be available for all temperature plugins, this can be put in the interface, but that would be an other pull request.


MFraters:
I actually moved part of the code to each of the temperature interfaces, with some part in a new features/utilities file.

PR #159, opened 2020-06-05
adress more warnings.

MFraters:
Found some more warnings which are triggered in travis, but not on my computer.


MFraters:
The python wrapper created by swig has some warnings, but there is not much I can do about that without getting rid of swig and writing my own wrapper. Gcc 4.7 also generates some -Wshadow warnings, but gcc 6+ doesn't, so I am trusting that.

MFraters:
This also fixes the coveralls problem by switching the coverage tester from trusty to bionic.

PR #161, opened 2020-06-07
Create lpo interface

MFraters:
This adds an interface for the Lattice Preferred Orientation (LPO) particle plugin for ASPECT. The basic idea is that the World Builder is able to return the 9 entries of the rotation matrix for each grain/crystal latttice in one vector and the size of the corresponding grain in the other vector.


MFraters:
This pull request adds the interface for grains (that is a list of grain sizes and a list of the rotation matrices of those grains) to the word builder. The code here increases the coverage percentage of the world builder, but codecov somehow things the tester is still running (the results can be viewed here: https://codecov.io/gh/GeodynamicWorldBuilder/WorldBuilder/tree/fc7d6c29531599131762e636de3decd603a4beee). 

The status of this new interface is experimental, so the wrapper functions for other languages are not provided yet. The reason for that is mostly that only a uniform plugin for the grains interface is available, which is mostly useful for testing purposes. In real world cases a random uniform and random distribution would be a minimum and the documentation has not been extended for this yet. These have to be added in a different pull request for which I will create a tracking issue.

This pull request contains a bit more history than normal, because I left the most important parts of the development history in for reference. Also please note that this pull request build on top of the many pull requests from the last ~1.5 weeks. Those were parts which where easy to split out.

The world builder App has been extended to return grains, but the visualizer has not.

PR #164, opened 2020-06-09
Add declare model entries function to parameters.

MFraters:
Still need to add a test for it.


MFraters:
All new lines are covered, but less lines in total. OSX warning was introduced in #162, so should be fixed in a different pull request.

PR #166, opened 2020-06-09
Modernize CmakeList

MFraters:
I had some issues with flags not being passed through correctly, and I stumbled on an article with some tips to modernize the usage of cmake. I don't think with this it will follow all modern cmake best practices yet, but it addresses some major points (using target more extensively instead of cmake variables for example).


MFraters:
This pull request raises the minimum cmake version from 2.6 to 2.8.4, because cmake versions lower than 2.8.4 do not support the working directory in add_test. This is still lower than the default cmake in ubuntu 12.04 which is an ubuntu version which is no longer supported. This is also still lower than the ASPECT minimum cmake version which is 2.8.12.

PR #195, opened 2020-06-20
make cmake cxx flags private and lower cmake version for target_compile_options.

MFraters:


MFraters:
This pull request now also allows for turn on `-Werror` during testing for specifically the library, app, visualization and unit tests, which I do in this pull request as well. The reason I could not do that before was that it would also have affected the code generated by swig, which, dependent on your swig version gives of warning with some of the compiler flags used for compiling the world builder.

PR #202, opened 2020-08-07
remove using namespace from include file

tjhei:
Remove "using namespace std" from features/interface.h as this puts many
symbols into the global namespace.

### DIFF
--- a/include/world_builder/features/interface.h
+++ b/include/world_builder/features/interface.h
@@ -27,8 +27,6 @@
 #include <world_builder/parameters.h>
 #include <world_builder/point.h>
 
-using namespace std;
-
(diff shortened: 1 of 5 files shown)


MFraters:
Thanks for the pull request! I think it makes sense to get std out of the global namespace. 

The tester has two problems. Can you add a using std to subducting_plate.cc? I will fix the other issue in a separate pull request, because it looks like swig has recently been changed on osx, and it now generates deprecated code somehow. Once that is fixed, I will ask you to rebase.

MFraters:
@tjhei The problem is fixed. If you can rebase, it should pass the tester

PR #216, opened 2020-12-03
Fix bug in subducting slab plate model

MFraters:
This fixes a bug which was introduced in #125. The ideas and comments in that pull request are I think still correct about using the depth, but for some reason the depth was never actually used. This pull request makes sure that the depth is actually used. Since the changes where made for version 0.3.0, that is the only version affected by this bug.


MFraters:
This pull request is ready to merge. The person who reported to problem has reported that they also concider the problem being fixed. One of the codecov testers fails, but that is just because there are less lines. So the effective coverage has not changed.

PR #222, opened 2021-03-24
add simple github tester.

MFraters:
TravisCI is not working anymore, so changing to github actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,184 @@
+name: Github CI
+
+on: [push, pull_request]
+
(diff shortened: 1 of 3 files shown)


MFraters:
This pull request  has added everything which is important to github actions (linux, windows, osx, release, debug, coverage, indenting checks). It currently only checks it for the most recent version of the operating systems and with all optional features installed. So it is less extensive then the travis CI tests used to be, but that can be added back later.

MFraters:
Also re-enables the appveyor simple_c_example test in release mode.

PR #223, opened 2021-03-24
fix coveralls in github actions.

MFraters:


MFraters:
Migrating coveralls to github actions turned out to be harder than expected, so I have removed it for now. This should not be an issue since there is still codecov which does exactly the same thing.

PR #224, opened 2021-04-23
allow interpolation choice per feature.

MFraters:
Currently there is only one global value which controls what interpolation is used. Although that may be a good default option, having the option for more fine grained control can be useful.


MFraters:
I disabled swig in the github actions windows test. See #225 for more info.

PR #237, opened 2021-05-18
WIP: Add option for contious distance to curved planes

### LINKED ISSUE #130: Improve subducting plate: continious subduction direction
The current subducting plate can be subdivided automatically in many parts based on a monotone spline. In some situations this still is not good enough though. 

It is possible to make the actual subduction direction also follow the curve of the monotone spline, which should improve lateral slab continuity quite a bit. 

Initial testing shows that it works well for subduction zones which curve outwards, but more testing needs to be done.

MFraters:
This resolves #130, by adding the option for continuous slabs and faults. To do this, the interpolation option `continuous monotone spline`. I am planning to disable the `max distance` interpolation parameter for this interpolation option, since it makes to sense here I think. It still need a proper cleaning of the code, but it is functional. I am mainly making this pull request to see if all the testers are fine with these changes.


MFraters:
I think this pull request is ready to go.

1. The benchmarks seems to show to me that there is not a significant difference for the old slabs in run time, which is good.
2. The coverage seems to be good enough. There are some weird lines missing, but I think they are incorrect. This doesn't mean that testing of this new feature should not be expended, because it will need to. But that can happen in a follow up pull request.
3. I re-eneabled coveralls (I got it working again), because codecov was showing some weird missing coverage. Coveralls is also showing some, but not all of them. Will need further investigation.
4. Will add benchmarks for the new feature in a follow up pull request.
5. I may want to prohibit point interpolation for the continuous monotone spline, but I will need to think about it a bit more to see if there are useful cases or not.
6. The documentation of the distance from curved planes function can also still be much improved, but I am not sure yet what the best way to do this. It is fundamentally a complicated function, so it may require some external documentation with pictures to be useful.

MFraters:
So I found some issues with the spherical case and resolved them (and I added new tests for them). I also refactored the code (again) to reuse as much of the non-contiuous code as possible. With some tweaks, the new code does not seem to have negatively affect performance (within the margin of error) according to the benchmark. I am leaving part of the commit history (which has been squashed a lot already) since I think I may be valuable to have some of the process of the development in here. 

Here is a picture for the all the different interpolation schemes for the spherical case. Fro the cartesian case see #130.
four_interplations_options_for_slabs_spherical_front

PR #238, opened 2021-05-22
Add contiuous benchmarking plus comment

MFraters:
This pull request adds two tings:

1. A benchmark system with a single benchmark (through `cbdr sample`) which is run by github actions
2. A report madeby `cbdr analyzer` as a comment in the pull request.

This should simplify finding out if there are new issues with the speed of the world builder, and if improvements are proposed, how much they actually speed up (if they actually do). The benchmark setup follows the advice by https://github.com/asayers/cbdr (and also uses the tools made available) and builds in the benchmark action both the current master branch and the pull request branch. They are then run for 120 second in random order (after warming up). The results are then analyzed and posted.


MFraters:
Results can be see in https://github.com/MFraters/WorldBuilder-1/pull/2. If the testers pass, I think it is good enough to merge. The number of samples is still a bit small and the caching doesn't seem to work as well as I hoped. The number of samples should increase by a lot when I make the pull request for making the visualizer a lot faster.

PR #242, opened 2021-05-23
Check if cbdr exits before checkout action.

MFraters:
The checkout action seem to destroy the cache, so trying a new approach by first checking whether there is a cache.


MFraters:
The cache usage for the rust part has been improved, but the cpp part seems to have issues with not rebuilding when not needed. I also increased to benchmark time to 90 seconds to see if the results are getting significantly more accurate or not (so technically not helping to speed it up).

PR #246, opened 2021-05-27
Extend benchmarks

MFraters:
The goal of this pull request is to both extend the current benchmarks to include compositions and add new benchmarks including the new continuous interpolation scheme. This also reorganizes the benchnarks so that analyze and show withing the tester the results early and makes it easier to add new benchmarks.


MFraters:
Additional changes: the benchmarks have been parallelized and are all 15 minutes long. This seems to be the time it takes to more or less reliable results. The commenting can only be tested after this is merged and may require further work.

Issue #211, opened 2020-09-28
Add the ability to load in files for use in a way compatible with and without mpi programs


MFraters:
Using ascii files or tomograhy data files can be useful for setting up models. It can also be useful for adding topography data for example (#132).

Loading files and reading them  is in itself not hard. One of the problems that, if naively implemented, MPI programs using the world builder will 1. load the file from disk on every MPI progress and 2. store one instance of the data in memory per mpi program. This can become problematic for very large/many files.

Issue 1 can be prevented by adding the option for the world builder to compile with MPI and get the MPI communicator, load the file in on processor 1 and than distribute it through the rest of the mpi processes. This is the most important issue since it would prevent a lot of io use. @gassmoeller adviced to look at the utlities function read_and_distribute_file_content in aspect (https://github.com/geodynamics/aspect/blob/master/source/utilities.cc#L959). The MPI communicator can be passed, but it should also be possible to get a default MPI communicator can be directly asked for since it is a global static variable. This will have to be investigated further.

I am not sure whether issue 2 can be solved with MPI since has a distributed memory models and not not shared memory model, but maybe there is a trick. But theoretically it should be possible to load in data once per node and spread a pointer. But I am not sure that would be worth the effort.

In summary, I think it should be restively straightforward to implement it in a way that prevents issue 1, while issue 2 might be a lot harder. 

Todo list:
- [ ] Implement a way to find mpi and enable it if present.
- [ ] Implement a way to get the MPI communicator if the parent program wants to communicate it, or using the default MPI communicator.
- [ ] Add utilities function to load files in create world and/or individual plugins.

gassmoeller:
> But theoretically it should be possible to load in data once per node and spread a pointer. But I am not sure that would be worth the effort.

This goes into mixed-parallel-shared memory parallelization and usually only becomes an issue with very large datasets, I would advise against going there unless you have a specific need for it. Another option is to look into data partitioning (only store what is needed on the current process), but that has its own set of problems (how do you know which part of the domain is needed on this process?). Most tomography models are small enough to simply store on every process.

MFraters:
Thanks for the comment, that makes sense.

PR #254, opened 2021-05-29
Add very basic mpi support

MFraters:
Add very basic MPI support to the world builder so that the world becomes aware of it's MPI rank. CMake will detect whether MPI is present and the world will auotomatically initialize MPI if it is not initalized when compiled with MPI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -184,6 +184,54 @@ jobs:
       run: ctest -VV -C ${{ matrix.build_type }}
       
 
+  test_linux_and_macos_mpi:
+    needs: test_indentation
+    strategy: 
+      fail-fast: false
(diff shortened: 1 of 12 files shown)


MFraters:
Besides adding the ability of the world builder to know what MPI rank it is, I have also expaneded the test set to test older versions of cmake. Compared to the first message in this pull request, I have changed that if MPI has not been initialized, it just assumes that the rank is zero. Throwing an error would also be an option, but that would create issues for some of the tests. 

Will still need to do some cleanup for the tester, but since it passes a the old and news tests and it compiles and runs still fine with aspect, I think it is almost good to go.

MFraters:
This resolves the first two items on the list in #211, and this will be useful for many other new features too.

PR #255, opened 2021-06-05
Refactor and fix spherical slabs

MFraters:
Looking in more detail at some spherical slabs with a lot of points close to gather I noticed some strange behavoir. It turned out that the 2d spherical distance computation was less accurate than I hoped. The pull request refactors large parts of the `distance_point_from_curved_planes` function with a focus on changing how the distance is computed, but is also has some changes on how the direction of the x axis is computed. 

This pull request will change results and therefore break tests. I will still need to do a lot of cleanup and to go through all the broken tests to confirm that the new solution is correct.


MFraters:
I think this pull request is now ready to merge. 

The accuracy is improved and the performance in the benchmarks is about 10% better (some even 15%). There is also a slight improvement in the coverage, and more importantly some extra tests which specifically target some of the changes made. Unfortunately some checks had to be turned off because they had been very specifically engineered to the previous system and it was hard to figure out why they where exactly failing without the geometric drawings at hand they where made with. This is also a lesson to improve the documentation of tests if analytic tests are added. 

Although the pull request could be cut up in multiple pull request (for example on on functionality and one on performance improvements), I think these changes make most sense together.

MFraters:
And here is a picture of a spherical slab made with continuous monotone spline defined through 107 points (representing the location cascadia trench). The colors show the distance from the left side of the slab (as seen from the view in the picture).
new_3d_spherical_slab_107points

PR #262, opened 2021-06-10
force disable windows python testing. 

MFraters:
This should have already been off, but due to sudden changes in the tester it suddenly seems to find a swig.


MFraters:
I also needed to force apt to update to make the tester more reliable on linux.

Issue #219, opened 2021-03-03
World builder consuming lot of time in my models


alarshi:
I am running 3d global mantle flow models based on a seismic tomography model. The plate boundaries are implemented as a compositional field, ```faults```, using the worldbuilder. However, my models are running very slow (more than 10 times slower) when using the world builder. Below is the log file for a model with and without the faults (notice the time taken to set up initial conditions) :
With faults:
![log_with_wb](https://user-images.githubusercontent.com/19296817/109813294-c73e8f00-7bf2-11eb-96dc-42842e199990.png)

Without faults:
![log_wo_wb](https://user-images.githubusercontent.com/19296817/109813320-d0c7f700-7bf2-11eb-87c5-d132ec459747.png)

I was wondering if there is a way to modify the worldbuilder implementation for this model, in order to reduce the time it takes to set up the initial conditions.

MFraters:
Hey @alarshi,

Thanks for you question! Without seeing the input file, I find it hard to see exactly why it is suddenly so much slower, but there are several things which could cause it to slow down that much.
1. Faults are very generic features, which can do the same things as slabs (have curved angles for example). Both faults and slabs are a lot more expensive than the other features due to this flexibility. If you just want to have a fault in a continental plate which goes straight down, you can just make it a continental plate and set it to a different compositional field. 
2. For both slabs and faults (and all the other features) you can set a `min depth` and `max depth`. Everything above and below these bounding values is not checked whether the point is in the slab/fault. I suspect setting the `max depth` to the deepest depth at which the faults are present (crust or lithosphere) will help a bit since the world builder then doesn't have to search in the mantle whether those points are in the fault, but probably not a huge amount since it already creates a bounding box based on the total length of the feature. But it is worth a try. 
3. If the two options above do not work for you, then depending on your model the time might be considerably reduced by not only limiting the search in depth but also in the other two dimensions. This may be particularly useful if you have a lot of short faults in your model. We could take the same approach with the automatic depth bounding box since we know the maximum length and thickness of the fault (and we might even be able to use the direction, although that can be a bit tricky). See for example: https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/75e14c23dbde578fcc8b968aa35bc91b31fd51b8/source/features/fault.cc#L489-L490 (it says `total_slab_length`, but it should be `total_fault_length` of course...)

[...]

PR #264, opened 2021-06-11
Changes to include a bounding box in computing fault composition.

alarshi:
This PR aims to address the issue: GeodynamicWorldBuilder/WorldBuilder#219, i.e. to reduce the time taken to assign fault composition using WB, particularly when a large number of faults are used. It introduces a function, bounding_polygon_contains_point () that returns true only when a point lies inside the bounding_polygon such that the function distance_point_from_curved_planes is not called for all the model points but only if the point lies inside the bounding_polygon. The work is still in progress because the performance is not improved as expected in my tests.


MFraters:
Hey Arushi,

Thanks for making this pull request! The code looks very clean and documented which is great. We may need to look at more detail at the test results later, but before we do that I have some general comments:

1. I think the functions `get_bounding_polygon` and `bounding_polygon_contains_point` are more generally useful than just for the fault (I am thinking of the slab). So I think it would be better to place them in `source/utilities.cc` and `include/world_builder/utilites.h`. 
2. Although a box is technically a polygon, you are actually computing a bounding box. Could you update the names of the functions to reflect that?
3. Since your goal is to improve performance, computing the bounding box can actually be done during the parse parameter phase. There is no need to compute the min and max element of a vector which doesn't change every call to the composition (or temperature or grains) functions.
4. Could you indent the code. This works exactly the same as in aspect.

Let me know if you have any further questions!

alarshi:
Thank you, Menno, for your suggestions. :) I incorporated them in the latest commit, however, I am yet to do more tests to evaluate how this branch is performing. I will keep you updated!

MFraters:
I am curious what the difference is going to be. 

Could you rebase the pull request to the newest branch? I have added a new benchmark which should be able to measure the difference.

alarshi:
Just did! 
I also increased the buffer zone, based on a test I was working on.

source/utilities.cc
```diff
@@ -212,6 +212,60 @@ namespace WorldBuilder
...
+        {
+          x_list[j] = point_list[j][0];
+          y_list[j] = point_list[j][1];
+        }
+
+      // Corner coordinates of the bounding polygon
+      std::vector<Point<2> > bounding_box (4, point);
+
+      // The bounding box currently is hard-coded to include buffer zone based on the spherical geometry.
+      const double bounding_buffer_zone = 2.* (buffer_around_fault) / (const_pi * 2. * 6371000.) ;
```
MFraters, line 234: There is no reason to compute this in this function since you already have a parameter for `buffer_around_fault`. I also think this can be more easily generalized for all spherical geometries since the calling function has access to both the `depth` and the `radius from the center`. Adding both will give you the radius of the sphere/earth.

Could you also add a bit of an explanation of what this is calculating?

MFraters (commented):
Great, and the benchmarks look very promising with just 2 faults which are far away! 
I don't see why some of the the test would be failing, but I can take a closer look at that later.

One more question, could you also add this for the slab? If you think that would be too much work that I can do that later in a follow up pull request, but I wanted to give you the opportunity to do it yourself :)

P.s. I made a mistake in extending one of the benchmark, which I fixed in #268, could you rebase again?

alarshi:
Hi Menno, I rebased and added the bounding box in subducting plate model. I was also getting a small warning in ```utilities.cc```, which I fixed. I had to do force rebase, hopefully, I didn't break anything.

alarshi:
> Out of curiosity, what was the warning? Was it related to this pull request or something separate?

I got this: 
```/home/asaxena/opt/WorldBuilder/source/utilities.cc:741:32: warning: unused variable ‘P1P2_norm’ [-Wunused-variable]```

**Comment by MFraters** [comment]
hmm, strange. I am not getting that warning. What compiler are you using? It is correct though. Could you put this change in an other pull request? I am asking this partly because it is a separate issue, but also if we can get that to merge quickly the all tests should start to run automatically for you without me requiring to approve it every time :) 

You can just remove the variable `P1P2_norm` and directly use `P1P2.norm()` if it is only used one time.

**Comment by alarshi** [comment]
Hi Menno, created a PR #269 for the warning.

**Comment by MFraters** [comment]
I did actually look into the tests and nearly all of them are very easy to fix when a special case for Cartesian is added. The buffer can't be computed during the parse parameter phase and needs to be computed and added during the compute composition phase, because that is the only place where the radius is known (with the radius from the point and the depth, as explained above). 

We may need to do a bit more thinking about the spherical cases and what a good boundary distance is. The correct distance could be computed by the intersection line of the sphere of the Earth and the spheres generated by the the points on the corners of the bounding box and the maximum length+thickness of the slab. Doing that exact calculation is probably too costly, but it is a good minimum bound to be aware of. Taking twice the length+thickness and converting that to spherical distance at the equator may be good enough.

**Comment by alarshi** [comment]
Hi Menno, 

In the recent commit, I added the check for coordinate system type in calculating ```bounding_box ()``` and also moved the buffer zone computation in ```get_composition``` phase. 

Regarding the buffer zone calculation, I agree that it needs to be more accurate for spherical geometry. 
I may be incorrect here, but I was also thinking that the buffer zone should contain information about the cell size. The way I am thinking is that if the mesh is too coarse, there might be a situation where a coordinate of the fault lies in the nearby cell but the bounding box  (```fault_thickness + fault_length```) is not large enough to include the cell size. I am sure, please let me know. :)

**Comment by MFraters** [comment]
Thanks for making the changes. I am actually surprised that with a x2 it works for the spherical case, because I still had one spherical test which gave me issues when I tried it. 

I am not sure if I understand you correctly, but I don't think the cell size of the mesh should play any role in this. Mostly because the idea behind the world builder is to set up a meshless mathematical representation of the model, which can be queried by the program which has a mesh. If the mesh that other program is too coarse to resolve the features defined in the input file that is not something the world builder can do anything about. The only thing the world builder can do something about is make sure that it produces the right output based on the input, which is mesh independent. So, the optimizations need to be always on the safe side.

So we need to think up a worst case scenario (I am thinking around the poles) and make sure that there is not difference between the normal and optimized version. 

Besides that, we also need to find out why these optimizations do not make a difference in your case (or has that changed?). I haven't gotten around to testing your input file myself yet, sorry about that :(

File: source/features/fault.cc [hunk 2]
```diff
@@ -488,8 +490,25 @@ namespace WorldBuilder
...
+      // the idea is to use a small buffer zone using the approximate length of fault, such that the box covers the fault
+      // extent in depth. For spherical system, we assume that the equivalent buffer is spherical distance computed at the equator.
+
+      if (world->parameters.coordinate_system->natural_coordinate_system() == CoordinateSystem::spherical)
+        buffer_zone = 2 * (maximum_fault_thickness + maximum_total_fault_length) / (const_pi * starting_radius* 2);
+      else if (world->parameters.coordinate_system->natural_coordinate_system() == CoordinateSystem::cartesian)
+        buffer_zone = 2 * (maximum_fault_thickness + maximum_total_fault_length);
+
+      std::vector<Point<2> >  bounding_box_coordinates = WorldBuilder::Utilities::get_bounding_box (
+                                                           reference_point, coordinates, original_number_of_coordinates, buffer_zone);
```
bangerth, line 506: This is something that doesn't depend on the current point this function is checking. It would be easiest if you could just do this computation once after you read in the details of the fault in `parse_entries()`, and then store the bounding box in a member variable.

source/utilities.cc
```diff
@@ -211,6 +211,57 @@ namespace WorldBuilder
...
+    bounding_box_contains_point (const std::vector<Point<2> > &point_list,
+                                 const Point<2> &point)
+    {
+      for (size_t i=0; i<point_list.size(); i++)
+        {
+          // The points in the point_list correspond to the get_bounding_polygon() output.
+          if (point[0] < point_list[2][0] && point[0] > point_list[0][0] && point[1] < point_list[1][1] && point[1] > point_list[0][1])
+            return true;
+          else
+            return false;
```
bangerth, line 259: This isn't right. You return either `true` or `false` here, for the very first point you look at. In other words, you ignore the variable `i` altogether. You might as well remove the loop above it in line 253 :-)

source/utilities.cc
```diff
@@ -211,6 +211,57 @@ namespace WorldBuilder
...
+      std::vector<double> y_list(original_number_of_coordinates,0.0);
+
+      for (size_t j=0; j<original_number_of_coordinates; ++j)
+        {
+          x_list[j] = point_list[j][0];
+          y_list[j] = point_list[j][1];
+        }
+
+      // Corner coordinates of the bounding polygon
+      std::vector<Point<2> > bounding_box (4, point);
```
bangerth, line 230: What is the meaning of `point` here? You will overwrite all elements of this array of four elements later on, so the value of `point` is unimportant. I don't think you actually need this function argument.

bangerth (commented):
This isn't quite right I think, but you're close to it. You are computing the bounding box as a list of four points,  but that's not how this is typically done, and I can't quite convince myself that what you are doing is right. (I do know that `bounding_box_contains_point()` is not right, see comment below :-) )

Typically, a bounding box just stores the min/max values for the x and y variables (or lat/long variables, if you want). You can think of this as the bottom left and top right point. So the bounding box should just consists of two points, not four, and checking whether a point is inside then becomes really simple. Take a look at how deal.II does this: 
https://github.com/dealii/dealii/blob/master/include/deal.II/base/bounding_box.h#L305
https://github.com/dealii/dealii/blob/master/source/base/bounding_box.cc#L21-L40
In your case, you can probably ignore the tolerance and just set it to zero.

alarshi:
Thank you @MFraters, @bangerth for your reviews  :)
I will update the PR after incorporating your suggestions. I can do some tests with the updated master, and keep you posted on the timing results!

source/utilities.cc
alarshi: Yes, I don't remember why I had the loop at all. :D Thank you!

source/utilities.cc
alarshi: Yes, the value of the point is not important. It was just to create the correct data type for bounding_box. Removed it now, thanks!

source/features/fault.cc
alarshi: I remember moving the calculation of ```bounding_box_coordinates()``` in ```parse_entries()``` stage based on @MFraters 's suggestion. However, the reason I moved it back into the ```get_composition``` stage is that I needed to use the ```starting_radius``` value for computing the buffer_zone, which I couldn't use in  ```parse_entries()```. Maybe there is a way to resolve this by modifying the approach for ```buffer_zone``` computation.

alarshi: I just realized that I can access the ```starting_radius``` value in the ```parse_entries()```. Please ignore my previous comment.

MFraters: The `starting_radius` is dependent on both the position and depth of the request. You only have them during the request itself (i.e. `get_composition`). I don't see how you could get those values during the `parse_entries` phase.

alarshi: You are right! I am sorry, I confused myself by using ```starting_depth```.

alarshi:
Hello @MFraters, @bangerth: I tried to incorporate your comments in the latest commit. I could not move the computation of ```bounding_box_coordinates``` in the ```parse_entries``` stage. 
I tried testing our global models with the master and this branch, the results are rather disappointing. I get an increased time for setting initial conditions with this branch by a factor of 3 compared to the master branch. I saw a speed-up with the earlier version of this branch, (commit dde325de) compared to the earlier head of master. Therefore, I suspect I messed up something in this commit, but I am not sure what. 
Another problem I am having is that I am missing a couple of faults (all at mid latitudes) using the ```bounding_box_coordinates``` defined currently. I can increase the ```buffer_zone```, but that is not a lasting solution. 
The model including the bounding box is at https://www.dropbox.com/sh/8ccwzn1wgh5dind/AADPRNoFOb4Av9ihlRBTmCHDa?dl=0 and the model run with the master is https://www.dropbox.com/sh/ebxe17w4kym6hk7/AADGAGL6wTuCheWenhSTcAFca?dl=0. The JSON file for the world_builder is in the folder with the bounding box model, ```geodynamic_world_builder_gem_all_faults.json```.
Thank you for looking into it!

source/features/fault.cc
bangerth: Is that because you are building 2d bounding boxes only for the specific depth you are currently at?

include/world_builder/utilities.h
```diff
@@ -65,8 +65,23 @@ namespace WorldBuilder
     double
     signed_distance_to_polygon(const std::vector<Point<2> > &point_list_,
                                const Point<2> &point_);
+    /**
+     * Takes the list of the fault coordinates and computes a bounding box
+     * based on the minimum and maximum x,y values in the list, and an
+     * additional buffer zone computed using the maximum fault length.
+     * Currenly, only implemented for the spherical coordinates.
```
bangerth, line 72:
```suggestion
     * Currenly, only implemented for spherical coordinates.
```

include/world_builder/utilities.h
```diff
@@ -65,8 +65,23 @@ namespace WorldBuilder
...
+     */
+    std::pair<Point<2>, Point<2> >
+    get_bounding_box (const std::vector<Point<2> > &point_list,
+                      std::size_t original_number_of_coordinates,
```
bangerth, line 76: what does this parameter correspond to?

source/features/fault.cc
```diff
@@ -477,8 +479,25 @@ namespace WorldBuilder
       // todo: explain
       const double starting_radius = natural_coordinate.get_depth_coordinate() + depth - starting_depth;
 
+      double buffer_zone = 0.;
+
+      // Here, we check if all the coordinate points describing a fault lie within a bounding box. To do this, we use min and
+      // max points for a fault and add a buffer zone surrounding it. Since the fault extends over a depth (i.e., fault lenght),
```
bangerth, line 485:
```suggestion
      // max points for a fault and add a buffer zone surrounding it. Since the fault extends over a depth (i.e., fault length),
```

source/features/fault.cc
```diff
@@ -477,8 +479,25 @@ namespace WorldBuilder
...
+      // the idea is to use a small buffer zone using the approximate length of fault, such that the box covers the fault
+      // extent in depth. For spherical system, we assume that the equivalent buffer is spherical distance computed at the equator.
+
+      if (world->parameters.coordinate_system->natural_coordinate_system() == CoordinateSystem::spherical)
+        buffer_zone = 2 * (maximum_fault_thickness + maximum_total_fault_length) / (const_pi * starting_radius* 2);
+      else if (world->parameters.coordinate_system->natural_coordinate_system() == CoordinateSystem::cartesian)
+        buffer_zone = 2 * (maximum_fault_thickness + maximum_total_fault_length);
```
bangerth, line 492: Can you explain why you add `maximum_total_fault_length` to the buffer zone, and why you multiply by two? The buffer zone should just be a distance around points so that the bounding box is around these points, but that buffer zone should be independent of where the points actually lie. (I'm assuming that `maximum_total_fault_length` is something that incorporated information about the points of the fault, right?)

include/world_builder/utilities.h
bangerth: I think you could drop this argument and just use `point_list.size()` everywhere

source/utilities.cc
```diff
@@ -207,6 +207,51 @@ namespace WorldBuilder
...
+      for (size_t j=0; j<original_number_of_coordinates; ++j)
+        {
+          x_list[j] = point_list[j][0];
+          y_list[j] = point_list[j][1];
+        }
+
+      // Initialize any point to create bounding_box, it will be overwritten later.
+      Point<2> reference_point(point_list[0]);
+      // Corner coordinates of the bounding polygon
+      std::pair<Point<2>, Point<2> > bounding_box (reference_point, reference_point);
```
bangerth, line 227: You should be able to just write it like this, and then don't need `reference_point` at all:
```suggestion
      std::pair<Point<2>, Point<2> > bounding_box;
```

bangerth (commented):
I have a couple of questions that I currently don't understand. If you have a second, think about these and if we can come to some conclusion, I can test this again!

include/world_builder/utilities.h
alarshi: Oh yes, thank you! :)

source/features/fault.cc
alarshi: So, the way I implemented the bounding_box is a 2D box that contains all the possible surface points for which the composition is calculated. (i.e., the last if condition in ln 499 in ```faults.cc: WorldBuilder::Utilities::bounding_box_contains_point (...)``` The condition for the depth is satisfied in ln 498 in the same if statement.
To calculate the bounding box, I use two inputs: ``` coordinates``` and a ```buffer_zone```. The```coordinates``` is the vector of 2D points that define the fault line. Worldbuilder creates the fault using the ```coordinates, fault thickness, dip angle, and fault_length```. Using the min and max points in ```coordinates```,  we create the two diagonal points of the box: lower left and upper right. But, since the thickness of fault is not incorporated in the ```coordinates```, we need to add that to the two extreme points. Additionally, if the faults are dipping, we need to search laterally so that the surface projection of the fault is also covered in the bounding_box. For any value of dip > 0, the surface projection would not extend beyond the ```maximum_total_fault_length```.  Therefore, I add the thickness and the fault to the existing surface coordinates. The factor of 2 is somewhat arbitrary to calculate the bounding lat/long coordinates in a spherical system (I need to remove it for cartesian). If the faults lie on a great circle, we don't need any factor, and as they go towards the pole, the factor should increase.  But, I need to discuss more with you and @MFraters  on how to correctly use this.

source/utilities.cc
alarshi: I tried that, but it complains ```no default constructor exists for class "std::pair<WorldBuilder::Point<2>, WorldBuilder::Point<2>>"```.

MFraters: That is because I would not know what the default constructor would need to be, since a point requires a coordinate system. Although I guess it could be `invalid`.

source/features/fault.cc
alarshi: No, it is general for all cases. I tried to explain the function in a comment reply below.

source/features/fault.cc
bangerth: @alarshi Thanks for the explanation -- although I have to admit that I understand only half of it because I'm not familiar with the way the WorldBuilder describes fault, and so I have no idea what many of the words mean :-)

I very frequently look for bugs by looking for things that appear in one expression but don't go together. In this particular case, it is "thickness" and "length". I got confused why one would want to add these together because I think of one as an "extent across the fault" and the other as an "extend along the fault", and these seem conceptually independent. I can think of a very short fault (very small length) that requires a large buffer zone because it is deep, thick, and with a small dip angle. I can also think of a very long fault that requires a small buffer zone because it is shallow, not very thick, and dips vertically. So it seems wrong that the length of the fault appears in the buffer zone width. But maybe I'm also misunderstanding what WB considers the "length" of a fault :-)

source/utilities.cc
bangerth: In #306, I initialize a pair of points this way:
```
std::pair<Point<2>, Point<2> > bounding_box =
{{
      -std::numeric_limits<double>::max(),
      -std::numeric_limits<double>::max(),
      -std::numeric_limits<double>::max(),
      cartesian
    },
    {
      +std::numeric_limits<double>::max(),
      +std::numeric_limits<double>::max(),
      +std::numeric_limits<double>::max(),
      cartesian
    }
  }
```

bangerth:
See also #306/#307 now.

source/features/fault.cc
alarshi: @bangerth: It is confusing because I was trying to make the bounding_box work for my case without thinking through all the details. You are right, the ```fault_length``` and ```thickness``` are perpendicular. My thinking was that if we have a vertical fault, it is sufficient that the buffer zone contains just the fault_thickness because the fault_length will be checked by the depth condition. On the other hand, a fault almost horizontal (dipping along the X-axis) would have surface coordinates depending on both thickness and the fault length, therefore, requiring a conservative estimate of buffer_zone containing both the values. 
I might be thinking it wrong, it would help if I could discuss this with you sometime today. I also want to discuss how to implement the BoundingBox class defined PR#306 for our case. :)

alarshi:
Hi @MFraters @bangerth: I figured out the problem with the missing faults (I was dividing the length by 2*pi instead of multiplying 2*pi)! The latest commit fixes that and also uses the #306 patch to calculate the bounding box in the ```parse_enteries()``` stage. I have currently implemented the bounding box for both the spherical and cartesian systems for the tester, but that needs to change based on #308. 
The time to setup initial conditions using the bounding box has reduced by a factor of ~ 5, and all the faults look good :) Let me know your suggestions on how to modify this PR so that it is useful for the WB.

source/features/fault.cc
```diff
@@ -325,6 +326,55 @@ namespace WorldBuilder
...
+        }
+
+      // Here, we compute the spherical bounding box using the two extreme points of the box containing all the surface
+      // coordinates and an additional buffer zone that accounts for the fault thickness and length. The first and second
+      // points correspond to the lower left and the upper right corners of the bounding box, respectively (see the
+      // documentation in include/bounding_box.h).
+      // For the spherical system, the buffer zone along the longitudal direction is calculated using the
+      // correponding latitude points.
+
+      const double radius_of_earth = 6371000.;
```
MFraters, line 347: let's talk about this. I don't like to have a constant in here. It would have my preference to have this at run time. I am curious how much of a performance difference it would make, since we can precompute everything except the radius of the earth. The other option would be to put it in the input file. There would be two options for that, either put it with the spherical coordinates or put it in a new section which contains optimization values. The downside of putting it in the spherical coordinates section is that it may conflict with the values provided by aspect and may confuse users. With a new section, we could make it clear that this should only be used for optimization and that it is the users responsibility to make sure it is consistent with the input file.

source/features/subducting_plate.cc
```diff
@@ -329,6 +330,57 @@ namespace WorldBuilder
...
+        }
+
+      // Here, we compute the spherical bounding box using the two extreme points of the box containing all the surface
+      // coordinates and an additional buffer zone that accounts for the fault thickness and length. The first and second
+      // points correspond to the lower left and the upper right corners of the bounding box, respectively (see the
+      // documentation in include/bounding_box.h).
+      // For the spherical system, the buffer zone along the longitudal direction is calculated using the
+      // correponding latitude points.
+
+      const double radius_of_earth = 6371000.;
```
MFraters, line 350: same here

include/world_builder/utilities.h
```diff
@@ -65,8 +65,6 @@ namespace WorldBuilder
     double
     signed_distance_to_polygon(const std::vector<Point<2> > &point_list_,
                                const Point<2> &point_);
-
```
MFraters, line 68: could you restore this file?

MFraters (commented):
A few comments, one we need to talk about. One other request, could you also add this check to the temperature and grains for both the slab and the fault?

source/features/fault.cc
alarshi: Thank you for the reviews and your help! :) I pushed the changes in the latest commit (described more in the comment below)

alarshi:
@MFraters @gassmoeller @bangerth : I incorporated your suggestions for pre-defining most variables in the parse-entries stage. Compared to the master, I get about 4x speed using this branch time to set up initial conditions. I discussed this with @MFraters  and know that this can be reduced further with more accurate computation of the bounding box (including the fault dips ), but this simple calculation still helps a lot with my models :)

source/features/fault.cc
```diff
@@ -348,7 +404,9 @@ namespace WorldBuilder
               );
 
       // todo: explain and check -starting_depth
-      if (depth <= maximum_depth && depth >= starting_depth && depth <= maximum_total_fault_length + maximum_fault_thickness)
+      if (surface_bounding_box.point_inside(Point<2>(position_in_natural_coordinates.get_surface_coordinates(),
+                                                     world->parameters.coordinate_system->natural_coordinate_system())) &&
+          depth <= maximum_depth && depth >= starting_depth && depth <= maximum_total_fault_length + maximum_fault_thickness)
```
MFraters, line 409: Can you turn this around? With `&&` it should directly return false as soon as one of the conditions is false. So doing the cheap depth checks should be cheaper.

source/features/fault.cc
```diff
@@ -340,6 +394,8 @@ namespace WorldBuilder
       // the starting radius is the distance from the bottom of the model to the surface.
       const double starting_radius = position_in_natural_coordinates.get_depth_coordinate() + depth - starting_depth;
 
+      BoundingBox<2> surface_bounding_box = get_bounding_box(position_in_natural_coordinates, depth);
```
MFraters, line 397: You can actually prevent this computing by putting it directly in the if statement. It is only used there, so that could save a lot of combination with the comment at the if statement.

MFraters: same for the other occurens

source/features/fault.cc
MFraters: same for the other occurrences

source/utilities.cc
```diff
@@ -206,7 +206,6 @@ namespace WorldBuilder
       return *std::min_element(distances.begin(),distances.end()) * sign;
     }
 
-
```
MFraters, line 209: could you remove this change?

MFraters (commented):
I think I found two more performance improvements, otherwise good to go. I am still a bit worried about the poles, but I am going to make a test now.

MFraters:
I convinced my self for now that the this is not going to add new issues for the poles in spherical coordinates, so I think it is good to go once you have addressed my last comments :)

source/features/fault.cc
alarshi: Got it, thank you :)

alarshi:
Just a quick update on the performance after the latest suggestions by @MFraters: the test model on this branch take ~ 40 s to set up initial conditions and the master takes about ~ 350 s. 
All the faults look okay and I am very happy with this speed-up. Thank you @MFraters @bangerth @gassmoeller for your help :)

MFraters (approved):
That is a great speedup! I think we can push it even a bit more if needed, but I think this pull request is ready to be merged. I think the changes and the involved work are large enough for a changelog entry. Could you add an entry? You could either put it under changed or make a new section called something like performance (or somewhere else if you think that would be a better fit).

Thanks for sticking through with this pull request!

alarshi:
@MFraters: I will do that! Thank you for all the help in this :)
 Do you want me to squash commits, once the PR is ready?

MFraters:
hmm, these are indeed a lot of commits for the amount of lines which are changed (although it was a lot of work on your side). Squashing it into three commits (one for the the fault, one for the slab and one for the changelog) would be appreciated.

MFraters:
ah, sorry, I wasn't clear. Could you add the changelog entry under the [unreleaded] heading instead of [0.4.0]?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - The plugin systems (temperature, composition and grains) and the distance_point_from_curved_planes function now all pass a precomputed NaturalCoordinate, besides just the cartesian position. It turns out that this can make a significant performance differce. \[Issue found and solution suggested by Wolfgang Bangerth, implemented and tested by Menno Fraters; 2021-07-03; [#300](github.com/GeodynamicWorldBuilder/WorldBuilder/pull/300) and [#219](github.com/GeodynamicWorldBuilder/WorldBuilder/issues/219)\]
+- Introduces a bounding box for searching the fault and the subducting plate model properties (temperature, composition, grains) in the lateral direction. This reduces the computation time by constraining the number of points for which the expensive distance_point_from_curved_planes function is called. \[Arushi Saxena; 2021-07-15; [#219] (https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/219)\]
```
MFraters, line 20:
```suggestion
- Introduces a bounding box for searching the fault and the subducting plate model properties (temperature, composition, grains) in the lateral direction. This reduces the computation time by constraining the number of points for which the expensive distance_point_from_curved_planes function is called. \[Arushi Saxena; 2021-07-15; [#219](https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/219)\]
```

MFraters:
The space between [] and () is preventing it from becoming a proper link in markdown. @bangerth, do you still want to have a look?

MFraters:
I talked to Wolfgang, and it is good to be merged now. Thanks again! :tada:

PR #266, opened 2021-06-15
Add benchmark which tests faults far away.

MFraters:
This adds a fault benchmark which also should be able to capture any performance improvements by pull request #264. This is done by having one fault in the region of interest and two outside the region of interest.


MFraters:
I also added the two extra slabs outside the box for the `Spherical slab interpolation simple CMS`. This has doubled the benchmark time, so I hope #264 brings it back :)

PR #267, opened 2021-06-15
small cleanup of subducting plate

MFraters:
Mostly to test the new benchmark introduced in #266, but the changes clean the code up as well.


MFraters:
It is clear that the accuracy of the slower tests has gone down a bit due to less runs in the 15 minute benchmark time. So I hope #264 brings the run time down again or otherwise the tests has to be simplified a bit or extend the run time.

PR #269, opened 2021-06-16
Fixed for a warning.

alarshi:
I was getting a small warning in ```source/utilities.cc```, this PR is just a fix to it. I am using GNU gcc 9.3.0 compiler.


MFraters (approved):
Thanks for fixing this!

PR #272, opened 2021-06-21
apply clang-tidy notes.

MFraters:
I am going through the comments provided by clang-tidy. The core check didn't result in anything (which is very nice), but some other checks have some useful  and some less useful tips. I am going through them by group, starting with performance.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -361,3 +361,60 @@ jobs:
       shell: bash
       run: ctest -VV -C ${{ matrix.build_type }}
 
+
+  linux-gcc_mpi:
+    needs: [test_indentation,coverage]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-18.04','ubuntu-16.04']
+        build_type: ['Debug']
+
+    runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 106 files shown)


MFraters:
I went through all the clang-tidy performance, modernize, misc and mpi (although nothing here) checks and I ended up applying all but `modernize-use-auto` and `modernize-use-trailing-return-type` notes since those changes are too large for now. They may still be applied after the release. It is nice to see that there actually is a large performance gain with with the performance checks and there seems to be a moderate performance gain with the modernize checks. It is unfortunate to see that the coverage has fallen a bit, but that is mostly due to using 
```
constructor()
= default;
```
instead of 
```
constructor()
{}
```
where the first one apparently counts as one line and the second one as two (covered) lines. So there is no real loss in coverage. I do still want to go through a few more to see if they would improve the release code.

MFraters:
A few more clang-tidy checks have been done and applied. I am considering doing the auto related fixes as well, but those changes are too big for the clean up before the release. I have also expended the testers with a ubuntu 16.04 and 18.04 tester to make sure the everything (still) works with older environments. I also took this opportunity to clean up an unused (and since c++11 obsolete) function and expand coverage.

Issue #274, opened 2021-06-22
Release GWB v0.4.0


MFraters:
This is the pull request to release the world builder version 0.4.0. 

Version update pull request:
- [X] Update version numbers in repository
- [X] Manually build manual and check it
- [X] Wait for Aspect approval
- [X] Merge pull request when Aspect has approved release for inclusion.

After version update pull request is merged:
- [X] Make separate release branch based on this pull request 
- [X] Release GWB 0.4.0 on github
- [X] Release GWB 0.4.0 on Zenodo
- [X] increase version number to 0.5.0-pre

MFraters:
Release of versio 0.4.0 is complete!

PR #277, opened 2021-06-23
Update version to v0.5.0 pre

MFraters:
Resloves #274 and finished the release of version 0.4.0 of the world builder.


MFraters:
The reason the benchmarks are failing is because the master branch has a different version from this branch (which is updating it). After this, the benchmarks should work again. Otherwise, good to go!

PR #282, opened 2021-06-24
Adding the option to compress vtu files

### LINKED ISSUE #265: Add zlib option for vtu output of world builder Visualizer
Paraview has the option to read vtu files with partial compression with zlib. Using this to (optionally) encode the files would be useful since it can potentially significantly decrease the file size.

MFraters:
This pull request is aimed to resolve #265, by adding the option to compress vtu output. The current approach is to switch from a homegrow vtu writer to a header only library called vtu11. Initial tests are promising with an about 7 time reduction in size for a 2d 2800x1800 model with 2 compositions. For very small models the old approach seem faster, but for larger models the (the one just mentioned seems as fast uncompressed and faster compression on. If this holds up, this could be explained by that this model is limited by the writing speed, not computation speed.

In the current state the tests will fail because the output is differently formatted (or even compressed), but it should be equivalent. I will also need to make the output options available to the user in the grid file.


MFraters:
I have done a test on a larger model and the conclusions seem to carry over. The run speed improvements are not as big (3min21s vs 3min36s), but the file is much smaller (287M vs 1,3G) and the load times in paraview are also a **lot** faster (10s vs 36s), while the mesh and the values on the mesh still look good for both 2d cartesian and the 3d chunk. An even larger version of a similair model is now 2.3G instead of 9.4G, so that is also a nice improvement. Furthermore, the benchmarks have all become significantly faster (the lowest difference (99.9 C.I) is between 2.9% and 13.3% faster).

In short, this seems to be the way to go and using zlib when available seem like a good default. It should still be possible to turn off both compression an binary for debugging purposes in the input file. I will need to do some cleanup, make the options available and update the tests, and add some more tests.

MFraters:
The windows tests now also finally seem to work. Somewhere along the way it is defining a macro for max in windows in an include somewhere. I have rearranged it to be the last include and undef both min an max macros. 

I made a lot of small unrelated fixes and changes along the way, which I will put in separate pull requests first.

MFraters:
I am now happy with the state of this pull request. I think this is a large improvement in the both the performance of using the world builder visualizer and the use in Paraview. 

This also means that all the benchmarks are faster because of less overhead of the visualizer. This means that more samples can be generated in the allotted time and that the results have a higher chance of being more precise.

The only issue I am left with is testing. The world builder only has continuous testing for the cartesian box (2d and 3d) and they are in the ASCII format in the new structure. This means world builder visualizer is generally very poorly tested, which is because all the testing efforts have been focused on the library itself (and the app because that is vital for the tester). Although I have tested the zlib compression version manually (also with a chunk), adding a tests to the test suite for continuous integration testing is more difficult since systems without zlib will give uncompressed output (which is a good default). I could create the cmake logic to deal with this, but I think the time is better spend rethinking the the visualizer testing. 

I had to make some changes to the vtu11 library to get it to work on all systems. I kept them in a separate commit and I will make pull request for them upstream.

PR #283, opened 2021-06-24
Improve matrix.org explanation 

MFraters:
Improve matrix.org explanation based on comments from @alarshi and focus more on the matrix room than the space.


alarshi:
Hi Menno! I just had a look over the changes, and I think everything looks perfect!

MFraters:
Thanks for taking a look at it!

PR #287, opened 2021-06-25
Enable minGW in windows tester.

MFraters:
the build system variable in the windows testers was not used, which resulted in going to the default Visual studio build system. This pull request fixes that.


MFraters:
MinGW builds fine, but all but one tests pass. That one test is only a small fraction off, but because I don't have access to numdiff it will fail. For now I will just disable the minGW (it was already not running before, but now it is explicit).

PR #289, opened 2021-06-25
Avoid map containers for parameter transfer

gassmoeller:
Hi @MFraters,
@jdannberg, @alarshi and I were looking into optimizing our world builder application and I noticed that more than 50% of the time in the world builder is spent on creating and accessing this map parameter. I know this is an intrusive change in your interface, but I think it would be worth considering. Generally a map is a very slow container, because it needs to dynamically allocate memory for every insertion, and even worse, for a map that uses strings as keys (like here) it needs to do multiple string comparisons every time you insert or access an entry. What you really want to hand over is a structure that is known at compile time, and for that a simple `struct` like I created here is completely sufficient and much faster. It can also be extended as necessary if later you want to hand over more parameters. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 - The World Builder Visualizer will now use zlib compression for vtu files by default. If zlib is not available binary output will be used. \[Menno Fraters; 2021-06-26; [#282](github.com/GeodynamicWorldBuilder/WorldBuilder/pull/282)\]
+- The return argument type of the distance_point_from_curved_planes function has been converted from a map to a struct, requiring a change in the plugin interfaces for 'fault_models' and 'subducting_plate_models', but significantly increasing the speed of the function and all functions that access its returned values. \[Rene Gassmoeller; 2021-06-27; [#289](github.com/GeodynamicWorldBuilder/WorldBuilder/issues/289)\]
 
 ## [0.4.0]
 ### Added
--- a/include/world_builder/utilities.h
+++ b/include/world_builder/utilities.h
@@ -262,8 +262,8 @@ namespace WorldBuilder
       double distance_from_plane;
 
       /**
-       * The distance between edge of the plane and
-       * the point if the point is projected onto the curved plane.
+       * The distance between the the start of the first segment (usually at
+       * the surface) to the provided point following the (curved) plane.
        */
       double distance_along_plane;
 
--- a/include/world_builder/utilities.h
+++ b/include/world_builder/utilities.h
@@ -249,6 +249,21 @@ namespace WorldBuilder
         std::vector<double> m_a, m_b, m_c, m_y;
     };
 
+    struct PointPlaneDistance
+    {
+      double distance_from_plane;
+      double distance_along_plane;
+      double fraction_of_section;
+      double fraction_of_segment;
+      size_t section;
+      size_t segment;
+      double average_angle;
+
+      // This is unrelated and should not be stored in here, but some
+      // plugins rely on this structure as temporary storage space.
+      double local_thickness;
+    };
+
     /**
      * Computes the distance of a point to a curved plane.
      * TODO: add more info on how this works/is implemented.
@@ -289,7 +304,7 @@ namespace WorldBuilder
      * the original number. Note that no whole numbers may be skiped. So for a list of 4 points,
      * {0,0.5,1,2} is allowed, but {0,2,3,4} is not.
      */
-    std::map<std::string,double> distance_point_from_curved_planes(const Point<3> &point,
+    PointPlaneDistance distance_point_from_curved_planes(const Point<3> &point,
                                                                    const Point<2> &reference_point,
                                                                    const std::vector<Point<2> > &point_list,
                                                                    const std::vector<std::vector<double> > &plane_segment_lengths,
--- a/include/world_builder/utilities.h
+++ b/include/world_builder/utilities.h
@@ -339,7 +339,7 @@ namespace WorldBuilder
      * extra coordinates automatically, and still reference the user provided coordinates by
      * the original number. Note that no whole numbers may be skiped. So for a list of 4 points,
      * {0,0.5,1,2} is allowed, but {0,2,3,4} is not.
-     * 
+     *
      * The function returns a struct that contains which segment and section of the curved
      * planes the point is closest to, what fraction of those segment and section lies before
      * the point (looking from the start of segment/section), the distance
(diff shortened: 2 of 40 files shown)


gassmoeller:
Oh and fixing up the unit tests will be a small nightmare, but I wanted to hear your opinion first before moving forward with that.

gassmoeller:
Looks like the benefits are slightly less for the github actions tests, but still :-)

include/world_builder/utilities.h
```diff
@@ -249,6 +249,21 @@ namespace WorldBuilder
...
+    {
+      double distance_from_plane;
+      double distance_along_plane;
+      double fraction_of_section;
+      double fraction_of_segment;
+      size_t section;
+      size_t segment;
+      double average_angle;
+
+      // This is unrelated and should not be stored in here, but some
```
MFraters, line 262: hmm, I see what you mean. Let's leave that for a follow up pull request.

include/world_builder/utilities.h
```diff
@@ -249,6 +249,21 @@ namespace WorldBuilder
         std::vector<double> m_a, m_b, m_c, m_y;
     };
 
+    struct PointPlaneDistance
```
MFraters, line 252: Could you write a bit of documentation above this? Something like: A structure to hold information about the the relationship between a point and a (curved) plane. 

Personally I am not really happy with the name of this struct and the function, since it has grown to more than just a holding distances, but I don't have a better option, so I am open to suggestions, otherwise this will work fine.

source/features/subducting_plate.cc
```diff
@@ -641,28 +641,28 @@ namespace WorldBuilder
...
-          std::map<std::string,double> distance_from_planes =
+          WorldBuilder::Utilities::PointPlaneDistance distance_from_planes =
             WorldBuilder::Utilities::distance_point_from_curved_planes(position,
                                                                        reference_point,
                                                                        coordinates,
                                                                        slab_segment_lengths,
                                                                        slab_segment_angles,
                                                                        starting_radius,
                                                                        this->world->parameters.coordinate_system,
-                                                                       false,
```
MFraters, line 652: Could you revert this change? (I am assuming it is a copy mistake)

source/features/subducting_plate.cc
```diff
@@ -509,28 +509,28 @@ namespace WorldBuilder
...
```
MFraters, line 520: Could you revert this change? (I am assuming it is a copy mistake)

source/features/subducting_plate.cc
```diff
@@ -380,28 +380,28 @@ namespace WorldBuilder
...
```
MFraters, line 391: Could you revert this change? (I am assuming it is a copy mistake)

include/world_builder/world.h
```diff
@@ -24,6 +24,7 @@
 
 #include <world_builder/parameters.h>
 #include <world_builder/grains.h>
+#include <world_builder/utilities.h>
```
MFraters, line 27: why are you adding this?

MFraters (commented):
Hey Rene,

This is great, thanks (and also to @jdannberg and @alarshi) for testing it and making the pull request! The benchmarks indeed show a great improvement.

This was one of the things in the back of my mind I wanted to look into to change, because I don't think there is a real advantage to using a map over a struct in this case, and structs are indeed supposed to be faster. I didn't realize that it would make such a large difference though.

I think most of the unit tests can be easily fixed with a few find and replace statements. I would be interested to hear if it is more work than that.

For the stability of the interfaces, this is only affecting the internal interfaces, so this will only affect people who make their own plugins for the world builder. Since I am currently not aware of anyone who have made their own (I know some who want to make some, but haven't started yet and will be contributed to the repository when they will be made), I do not think this should be an issue. Furthermore, and the world builder is still below version 1.0, and a new version has been released just a few days ago. So I think this is the perfect time to make this kind of changes so that we have a year to test them :) Also, the fix for existing plugins would be trivial. 
So in short, if this would affect external interfaces I would be much more cautious and want at least a depreciation mechanism, but since it is only internals and the project is below version 1.0 I think it is good to make these changes.

So, thanks again for making this pull request, is saves me work ;) 

I do have a few comments and may have a few more after the tests are fixed, but they should be easy enough to fix.

gassmoeller:
What do you think about this version?

I had to include the utilities header file to let plugins know what `PointDistanceFromCurvedPlanes` is, however now I moved it to only the interfaces that actually need it.

include/world_builder/utilities.h
```diff
@@ -249,6 +249,57 @@ namespace WorldBuilder
...
+    struct PointDistanceFromCurvedPlanes
+    {
+      /**
+       * The shortest distance between point and plane.
+       */
+      double distance_from_plane;
+
+      /**
+       * The distance between edge of the plane and
+       * the point if the point is projected onto the curved plane.
```
MFraters, line 266: I do not understand what you mean with edge of the plane here, so I am not sure whether this description is correct. This should be something like: 
```
The distance between the the start of the first segment (usually at the surface) to point a point on the (curved) plane to the provided point following the (curved) plane.

File: source/features/fault.cc [hunk 8]
```diff
@@ -508,14 +508,14 @@ namespace WorldBuilder
...
-          const double distance_along_plane = distance_from_planes["distanceAlongPlane"];
-          const double section_fraction = distance_from_planes["sectionFraction"];
-          const size_t current_section = static_cast<size_t>(std::floor(one_dimensional_coordinates[static_cast<size_t>(distance_from_planes["section"])]));
+          const double distance_from_plane = distance_from_planes.distance_from_plane;
+          const double distance_along_plane = distance_from_planes.distance_along_plane;
+          const double section_fraction = distance_from_planes.fraction_of_section;
+          const size_t current_section = static_cast<size_t>(std::floor(one_dimensional_coordinates[distance_from_planes.section]));
           const size_t next_section = current_section + 1;
-          const size_t current_segment = static_cast<size_t>(distance_from_planes["segment"]); // the original value was a unsigned int, turning it back.
+          const size_t current_segment = distance_from_planes.segment; // the original value was a unsigned in, converting it back.
```
MFraters, line 516: could you remove the comment behind the line? It is no longer relevant.

source/features/fault.cc
```diff
@@ -642,14 +642,14 @@ namespace WorldBuilder
...
```
MFraters, line 650: same here

MFraters (commented):
> What do you think about this version?

Great! I only have a few minor comments, and then it is ready to merge.

> I had to include the utilities header file to let plugins know what `PointDistanceFromCurvedPlanes` is, however now I moved it to only the interfaces that actually need it.

hmm, `PointDistanceFromCurvedPlanes` is way too long of a name and doesn't really cover what is in the struct. But it lines up with the function name, which is fine for now. It can be changed together with the function name as soon as a better name for both has been invented ;)

MFraters:
Could you also add a line in the changelog? I think under a subheading changed would be appropriate.

gassmoeller:
Ok I fixed the comments and added a changelog. Please take a look how I addressed your suggestion for the documentation, your comment was not quite clear to me, and I do not quite understand what this variable is even now, so better take another look.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - In the fault temperature model linear, the entiry top temperature is changed to center temperature and the entry bottom temperature is changed to side temperature, since this represents the actual sides more accurately. \[Menno Fraters; 2021-07-09; [#260](github.com/GeodynamicWorldBuilder/WorldBuilder/issues/260)\]
 - A large overhoal of the distance_point_from_curved_planes function improving the accuracy in spherical coordinates. This may slighty change the results, but it should be an improvement both in accuracy and performance. Also makes available a coordinate system independent distrance function and some fast trigonometric functions. \[Menno Fraters; 2021-06-11; [#255](github.com/GeodynamicWorldBuilder/WorldBuilder/issues/255)\]
+- The return argument type of the distance_point_from_curved_planes function has been converted from a map to a struct, requiring a change in the plugin interfaces for 'fault_models' and 'subducting_plate_models', but significantly increasing the speed of the function and all functions that access its returned values. \[Rene Gassmoeller; 2021-06-27; [#289](github.com/GeodynamicWorldBuilder/WorldBuilder/issues/289)\]
```
MFraters, line 31: This should go in the unreleased section. If you rebase to the newest master, there will be some entries in there already.

MFraters (approved):
Just one small comment about the changelog. The documentation is at least correct now, although I agree it is not clear yet. I think it will be easier if I update the documentation of this class in a follow up pull request and then ask you whether is has become more clear.

gassmoeller:
Like this?

MFraters:
> Like this?

Yes, great. Thanks for fixing this!

PR #290, opened 2021-06-25
Use DEBUG to indicate debug mode

gassmoeller:
I will get back to #289 soon, but first a question: Is there a reason that world builder uses NDEBUG to distinguish between debug and release mode? deal.II and ASPECT uses DEBUG, which leads to the fact that currently in aspect all `WBAssert` calls are executed even in release mode. I have seen that RapidJSON seems to use NDEBUG, but it matters in neither case, because world builder does not seem to set DEBUG or NDEBUG at the moment. The change I did below saves about 20-25% for my example model (mostly because operations on world builder points become much faster).

If you agree that this makes sense, I can change the remaining occurences of NDEBUG into DEBUG (all in the tests folder). In order for the world builder itself to distinguish between debug and release mode the world builder build system would need to set this variable if necessary. This fix will only affect WB if it is compiled by ASPECT.

### DIFF
--- a/include/world_builder/assert.h
+++ b/include/world_builder/assert.h
@@ -26,7 +26,7 @@
 
 namespace WorldBuilder
 {
-#ifndef NDEBUG
+#ifdef DEBUG
 #   define WBAssert(condition, message) \
   do { \
       if (! (condition)) { \
@@ -37,7 +37,7 @@ namespace WorldBuilder
         } \
     } while (false)
 #else
-#   define WBAssert(condition, message) do { } while (false)
+#   define WBAssert(condition, message) {}
 #endif
 
 #   define WBAssertThrow(condition, message) \


MFraters:
>  Is there a reason that world builder uses NDEBUG to distinguish between debug and release mode?

hmm, to look it up I did a git blame, but unfortunately this is the second commit (the first one just adds a license), which actually used the boost property tree, so it is not because of rapidjason, but otherwise not much help there. 

I had to look it up, but I think the reason I am using NDEBUG is because it is more standard than DEBUG (see for example https://en.cppreference.com/w/cpp/error/assert and https://stackoverflow.com/questions/2290509/debug-vs-ndebug). Cmake should also set the flag `-DNDEBUG` when in release mode. If I look at `ccmake` in the world builder, the the following flags:
```
 CMAKE_CXX_FLAGS_DEBUG            -g  
 CMAKE_CXX_FLAGS_MINSIZEREL       -Os -DNDEBUG
 CMAKE_CXX_FLAGS_RELEASE          -O3 -DNDEBUG
 CMAKE_CXX_FLAGS_RELWITHDEBINFO   -O2 -g -DNDEBUG
```
Which should get me the correct behavior. I can't quickly find where aspect is setting DEBUG, because Aspect doesn't really seem to use these cmake parameters (both debug and release are empty for me):
```
 CMAKE_CXX_FLAGS_DEBUG
 CMAKE_CXX_FLAGS_MINSIZEREL       -Os -DNDEBUG
 CMAKE_CXX_FLAGS_RELEASE
 CMAKE_CXX_FLAGS_RELWITHDEBINFO   -O2 -g -DNDEBUG
```
Looking at the effective flags in aspect (through `cmake .. -DASPECT_ADDITIONAL_CXX_FLAGS=-ignore_me_just_testing` I get the following: 
```
-- Appending ASPECT_ADDITIONAL_CXX_FLAGS: '-ignore_me_just_testing':
--   DEAL_II_CXX_FLAGS_DEBUG: -O0 -ggdb -Wa,--compress-debug-sections -ignore_me_just_testing
--   DEAL_II_CXX_FLAGS_RELEASE: -O2 -funroll-loops -funroll-all-loops -fstrict-aliasing -Wno-unused-local-typedefs -ignore_me_just_testing
```

I am assuming now that the debug macro is defined somewhere in a header in deal.ii, but searching `DEBUG` in the deal.ii directly is a lot of stuff to go through :P Do you know why Aspect and DEAL.II are using DEBUG instead of NDEBUG, and where it is setting it?

I certainly agree that this is a problem which needs to be fixed soon, either here or in aspect, but I am not sure what the right approach is.

gassmoeller:
You are right, this is something that should be fixed in ASPECT or deal.II, it is up to the World builder to decide how to activate assertions, and NDEBUG seems a standard choice. I dont know if there is a reason why deal.II (and therefore aspect) does it differently. I will open separate discussions there, but will keep this open for now for reference.

gassmoeller:
Ok, we have a fix in ASPECT. Have not heard back from deal.II yet, but it does not matter, since one of the two setting NDEBUG properly is enough. I will close this.

PR #293, opened 2021-06-27
clang-tidy option in cmake an fix for cpp 14 

MFraters:
This adds an option to run the clang-tidy check when compiling. It is off by default, because it makes compiling slower and is only useful when developing (could turn it on in debug mode in the future). I also am rerunning and fixing a select set of clang-tidy checks clang-tidy now that I am using cpp14. I am not sure whether all changes have to do with c++14 or I just missed them before.

I am also considering adding a test whether there are any issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,23 @@ set (WORLD_BUILDER_SOURCE_DIR ${PROJECT_SOURCE_DIR})
 # generate version.cc
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/source/config.cc.in" "${CMAKE_BINARY_DIR}/source/config.cc" @ONLY)
 
+# determine whether to use clang-tidy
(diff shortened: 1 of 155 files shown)


MFraters:
I think these changes are sensible and considerably clean up the code. I also think they best to do right after the release, so that future development for new contributors will hopefully be easier. Although I do not think will make a real git merge conflict, I will wait on merging for #289 since it is nearly almost ready to merge and it includes some header files which will be changed. I don't think I need to wait for #264, since the changes there are not close the the changes in this pull request.

In summary, this modernizes the code for cpp14 in several areas and introduces a few readability improvements. They can be automatically checked and even fixed though cmake, but I don't think I will enforce this strictly for the time being. There is a will also be a follow up pull request to clean the code up even more (remove unused headers and fix some clang warnings (I will also need to add a non-mac clang tester, they are really different in their warnings apparently))

PR #296, opened 2021-06-28
Clang tidy headers fixed

MFraters:
This pull request changes two things: 

1. remove `readability-avoid-const-params-in-decls` from the list, because I think that actually decrease readability, and 
2. Apply fixes from warnings when clang-tidy is applied to the world builder headers. This includes the manual fixes to make functions final.


MFraters:
I am marking all the final plugins as `override final`. For most compilers just `final` is enough, but the gcc (7.5) compiler in Ubuntu 18.04 really wants the override before the final.

PR #300, opened 2021-07-03
precompute natural coodinates.

MFraters:
Found in #219. Moves the computation of the natural coordinates outside the feature loop and pass it as a reference. This will also contain a commit later to pass the natural coordinates by reference to the distance from curved planes function.

Although it currently looks kinda inconsistent to have a value called point and one called natural_coordinate, I don't think it is a good idea to make them consistent in this pull request, and focus here on the functional change. 

edit: although I am open to changing the names in a last seperate commit here or in a new pull request.


MFraters:
I think this pull request is also good to go. Thanks for suggesting this @bangerth!
This builds on the improvements made in #301. The benchmarks indicate that it either makes no difference or is significantly faster. My tests also show that it can be sigificantly faster (more than indicated by these benchmarks). I don't like the long namespace identifier `WorldBuilder::Utilities::NaturalCoordinate`, but that is just needed unless I want to define a using, which will conflict with the features utilities namespace making it even messier.

Also related to the work done in #264

MFraters:
It is important to note that this pull request doesn't affect individual plugins at all, just plugin systems. Individual plugins still only get the Cartesian point.

bangerth (commented):
Yes, that's exactly what I had in mind. Nice to see that it makes a difference!

Issue #302, opened 2021-07-04
cartesian dot product used in spherical coordate system


MFraters:
@bangerth mentioned  in #301 that the dot product as defined in the point class in only valid for cartesian coordiante systems. I agree that is correct, but I have been using it in 2d spherical system as well as an approximation. This approximation seems to be working well enough for the cases where I had tested it, but I think it will only hold if the point are not too far away from each other. 

I am currently viewing fixing this issue as a way to gain some accuracy, but I am happy to be proven wrong.

MFraters:
With 2D I mean that I omit the the radius, so the 2D point contains longitude and latitude, so unit wise it is fine (both angles in spherical coordinate system).

bangerth:
*Locally* every smooth manifold is Cartesian, so if the two points are close to each other, then your approach works.

But take a point on the equator, `x=(phi=0, psi=0)`, and *any* other point `y`. In your scheme, their dot product is zero suggesting that the two vectors are orthogonal when that is clearly not the case. In fact, `x*x=0` even though the point is not the origin, just a point on the sphere. A similar paradoxon appears if you choose `x=(phi=0,psi=pi/2)`, a point on the equator. Then choose `y=(phi=pi/2,psi=0)`, the north pole. You have `x*y=0` as expected because the two points are perpendicular to each other. But `y=(phi=pi/2, psi=pi/2)` is *also* the north pole and `x*y=pi^2/4`, suggesting that the two points are not perpendicular to each other when clearly they are.

MFraters:
Thanks for the explanation. 

Even though this is only used to define the details of the trench and fault lines, which are inherently small scale, the corner cases are worse that I expected. This approach is actually only in use in the old/simple system. The new continuous slab and faults use a completely different way of getting the information (actually computing the distance at all points which form the line of the line and three points in between, and then doing a search in the area with the smallest distance to get a better resolution). This new method also has the advantage of making the slabs and faults continuous. 

I think everyone should switch to the new approach anyway, but it is slightly more expensive if I remember correctly and I am not sure how much that difference is affected by all the changes and optimalizations after that pull request. The following two benchmark results from the last merged pull request should be a fail comparison I think:

Benchmark | Master | Feature | Difference (99.9% CI)
-- | -- | -- | --
Slab interpolation curved simple none | 1.422 ± 0.039 (s=293) | 1.421 ± 0.043 (s=343) | -0.8% .. +0.7%
[...]

PR #304, opened 2021-07-04
WIP: Remove non contiuous option from distance point from curved planes

MFraters:
This is a first attempt to address issue #302.

 This will change the result for the old interpolation schemes and therefore break some tests, but in my opinion the new result is closer to what a use would expect anyway. I see no reason why a user would want to have a slab or fault which is discontiuous at the points which define the location of the slab/fault. If a user wants a discontiuous areas in a slab, they can still create it explicitly.

So if this pull request is implemented, all the slabs will be continuous. At the same time allows this allows for some (hopefully significant) optimizations in the interpolation scheme.

### DIFF
--- a/include/world_builder/point.h
+++ b/include/world_builder/point.h
@@ -260,7 +260,10 @@ namespace WorldBuilder
        * function without asin and sqrt is returned.
        */
       double
-      cheap_relative_distance(const Point<dim> &two) const;
+      cheap_relative_distance_spherical(const Point<dim> &two) const;
+
+      double
+      cheap_relative_distance_cartesian(const Point<dim> &two) const;
 
       /**
        * return the internal array which stores the point data.
--- a/include/world_builder/point.h
+++ b/include/world_builder/point.h
@@ -369,10 +369,10 @@ namespace WorldBuilder
      */
     inline double fast_sin_d(const double angle)
     {
-        WBAssert(angle >= 0 && angle <= const_pi, 
-        "You are tring to use the fast sin function in an invalid range. "
-        "The fast sin function can only be used for values between 0 and pi. "
-        "The provided angle is " << angle);
+      WBAssert(angle >= 0 && angle <= const_pi,
+               "You are tring to use the fast sin function in an invalid range. "
+               "The fast sin function can only be used for values between 0 and pi. "
+               "The provided angle is " << angle);
       constexpr double A = 4.0/(const_pi *const_pi);
       constexpr double oneminPmin = 1.-0.1952403377008734-0.01915214119105392;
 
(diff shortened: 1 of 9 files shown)


include/world_builder/point.h
```diff
@@ -381,25 +388,37 @@ namespace WorldBuilder
...
     {
-      const double angle = (raw_angle > -const_pi && raw_angle < const_pi)
-                           ?
-                           raw_angle
-                           :
-                           FT::fmod(raw_angle + std::copysign(const_pi,raw_angle), const_pi * 2.0) - std::copysign(const_pi,raw_angle);
-
-      if (angle >= 0)
+      WBAssert(angle >= -const_pi && angle <= const_pi,
+               "You are tring to use the fast sin function in an invalid range. "
```
bangerth, line 397:
```suggestion
               "You are trying to use the fast sin function in an invalid range. "
```

MFraters:
Thanks for taking a look at the code Wolfgang! I fixed the spelling mistake :)

The pull request is currently not ready yet. I am slowly getting closer to the performance I would like (at least no big performance loss and still some values to tune), but to do that I had to cut a lot of corners some which I some certainly will have to reverse. I am happy though that I managed to get the `Spherical fault interpolation simple none` from about +135% to slightly negative :)

Part of the solution for that particular example is to actually do something similar to #264. This will in no way replace the functionality in that pull request, because it is  has expensive code before it and is using an expensive functions to do it. At the current state my test for @alarshi 's model (`./bin/WorldBuilderVisualization geodynamic_world_builder_faults.json ../test/visualization/spherical_subducting_plate_gridfile.grid`) went from 2min and 4 sec(14403M instructions/cycles according to callgrind) to 1 min and 17 sec. (9595M instructions/cycles).

Some of the speed ups would also have been possible with the old method, although not all. The main questions I am having is whether to still allow the old method for it's speed. That would mean that time needs to be invested in fixing the issue discussed in #302, which will most likely incur a small but significant cost on the performance. I could leave the old method available through a separate function, without the option to interpolate coordinates.

This pull request still requires a bit of work and (a lot of) cleaning up and testing, but I am now happy with the direction it is going.

bangerth:
I have trouble identifying what it is you wanted me to look at. Can you point me at a specific commit (or just move unrelated commits into separate PRs)? I'd be happy to look!

MFraters:
ah, sorry. I haven't gotten around to clean it up yet. I will certainly ping you when it is ready to take a look. I think a agree with what you said before, and I will focus this pull request on the functional change of removing the old option in favor of the new option. And each optimization will have it's own pull request.

PR #305, opened 2021-07-06
fix bug in calculation of maximum fault and slab thickness.

MFraters:
I found this while working on #304. The maximum thickness was defined as the total length. This should not change any results, but may speedup the benchmarks.

### DIFF
--- a/source/features/fault.cc
+++ b/source/features/fault.cc
@@ -316,11 +316,12 @@ namespace WorldBuilder
               local_total_fault_length += segment_vector[i][j].value_length;
 
               fault_segment_thickness[i][j] = segment_vector[i][j].value_thickness;
+              maximum_fault_thickness = std::max(maximum_fault_thickness, fault_segment_thickness[i][j][0]);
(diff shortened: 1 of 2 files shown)


MFraters:
In hindsight it makes sense that it doesn't effect any of the benchmarks, since they all have slabs/faults going to more or less the bottom of the box... Might be worth considering adding a benchmark which has a box which goes much deeper than the slabs/faults in it.

bangerth:
@MFraters Interesting. Is this then also something that seems wrong to you here? https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/264/files/f67d77056ab1409fe4ed9b0e4501dfb4fe6860c1#diff-2062fc3ee4671a7b02b0c338abb50f5d4e117226be9b310863a8a0cd286bceb9 I had trouble understanding why the length appears there in the buffer zone, but maybe I'm also just not understanding how "length" is defined here.

Either way, there is also this line that also makes the "length" appear in a "depth" comparison: https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/master/source/features/fault.cc#L482

PR #306, opened 2021-07-06
Add a BoundingBox class.

bangerth:
This patch adds an implementation of a `BoundingBox` class. I'm not sure I've followed your style guides, but you will tell me if not. 

I'm not using this at the moment, but wanted to get this out of the way for follow-up patches that actually make use of it.

### DIFF
--- a/include/world_builder/bounding_box.h
+++ b/include/world_builder/bounding_box.h
@@ -343,6 +343,11 @@ namespace WorldBuilder
   BoundingBox<spacedim>::point_inside(const Point<spacedim> &p,
                                       const double tolerance) const
   {
+    WBAssert(boundary_points.first.get_coordinate_system() == p.get_coordinate_system(),
+             "Cannot compare two points which represent different coordinate systems.");
+    WBAssert(boundary_points.second.get_coordinate_system() == p.get_coordinate_system(),
+             "Cannot compare two points which represent different coordinate systems.");
+
     for (unsigned int i = 0; i < spacedim; ++i)
       {
         // Bottom left-top right convention: the point is outside if it's smaller
--- a/include/world_builder/bounding_box.h
+++ b/include/world_builder/bounding_box.h
@@ -98,10 +98,11 @@ namespace WorldBuilder
   {
     public:
       /**
-       * Standard constructor. Creates an object that corresponds to an empty box,
-       * i.e. a degenerate box with both points being the origin.
+       * Standard constructor. Creates an object that corresponds to a
+       * box that corresponds to the entire space, i.e. a degenerate
+       * box with end points at minus and plus infinity.
        */
-      BoundingBox() = default;
+      BoundingBox();
 
       /**
        * Standard constructor for non-empty boxes: it uses a pair of points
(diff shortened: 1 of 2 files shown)


MFraters:
Thanks for the pull request. How much did you change the deal.ii code? If you hardly changed it and are not planning to change it further, it might make more sense to make a separate direction in the include folder for deal.ii and a deal.ii or WBDeal.ii namespace. Otherwise it would probably better fit in the utility class and namespace.

bangerth:
I removed half a dozen member functions that we won't need, and switched `Assert` to `WBAssert`, but other than that made no changes to the actual code. I don't plan on making any additional changes.

If you tell me where you want the code, I can put it there.

MFraters:
for the location of the file, let's just leave it here for now. I am mostly worried that it will start to fill up the include folder, but let's worry about that when it becomes close to a real problem or when I had more time to think about it.

I don't think the current code will actually work, since the point class is not defined. So you would either need to also import part of the deal.ii point (creating incompatibilities) or use the WB point. The last one would have my preference, but the WB point class needs to have a coordinate system defined. 

Could you also add a unit test to make sure it actually compiles and works? You can do that by making a new file in the `tests/unit_tests/` directory. It uses catch2, like aspect does, so it should be familiar :)

include/world_builder/bounding_box.h
```diff
@@ -0,0 +1,446 @@
...
+       * iterators to Point<spacedim> elements.
+       */
+      template <class Container>
+      BoundingBox(const Container &points);
+
+      /**
+       * Return a reference to the boundary_points
+       */
+      std::pair<Point<spacedim>, Point<spacedim>> &
+                                               get_boundary_points();
```
MFraters, line 127: This alignment is freaking me of a bit. Could you either get it on one line or alligned with the beginning of the line. apparently astyle doesn't fix this.

include/world_builder/bounding_box.h
```diff
@@ -0,0 +1,446 @@
...
+
+      /**
+       * Return a const reference to the boundary_points
+       */
+      const std::pair<Point<spacedim>, Point<spacedim>> &
+                                                     get_boundary_points() const;
```
MFraters, line 133: same here.

MFraters (commented):
Is there a reason why this is all header only? I generally prefer a separate header and source file, since it works better with the coverage and clang tidy checks.

bangerth:
Now seems to work, and also has a test :-)

bangerth:
As for header-only: The class really only has a minimal set of functions. It's really just a marginal abstraction over `std::pair<Point,Point>` with a function `point_inside()`.

MFraters (commented):
Thanks for writing the tests! I think the functionality can be certainly useful and I like the simple interface. It looks mostly good to go. I just have two more questions:

1. This only works for Cartesian right? If so, could you add an assert for spherical points?
2. could you rebase to the newest master?

bangerth:
It should work in whatever coordinate system you are working on, depending on how exactly you define "inside the bounding box". The way the `point_inside()` function defines it is that each coordinate is within lower and upper bounds, whatever they may represent. There is no check for periodicity. But you can use this to check whether a point is within a certain `min_latitude...max_latitude` and `min_longitude...max_longitude` range.

So I'm not sure whether you actually want the assertion you mention. I'll leave that up to you. I did add an assertion that the point given to `point_inside()` must use the same coordinate system as the points that describe the bounding box.

bangerth:
I rebased.

MFraters (approved):
Oke, if it is fine with other coordinate systems, than the new assert you just added is good enough. I will take one more look at it tonight, but I think it is ready to be merged.

MFraters:
Congrats and thanks for making your first pull request to the world builder!

bangerth:
Ha, there's always a first time! :-)

PR #308, opened 2021-07-07
[WIP] Use bounding boxes

bangerth:
I need to move on to other stuff, but here is the idea of using the bounding box stuff from #306 (only look at the last commit, the first ones are duplicates from #306).

I haven't (and can't) implemented the code that actually determines the bounding box. I have no idea how you represent faults here, but maybe that's something one of you can help with?


bangerth:
OK, now only contains the new part, rebased on top of #306. There is no functionality here, as I need help for that :-)

MFraters:
While writing down an explanation, I now realize that this approach might not help. The problem is that the the size (and depth) of the domain is not defined anywhere in the input file. The reason for that is that information is simply not needed to make the model, because al the information which is needed to uniquely define the relevant model domain (which are the location of the point and the depth at which that point is) are provided at the time the request is made. Adding more information to the input files would in this case lead to inconsistencies. Let's talk about this tomorrow. I left the explanation below, since I still think it might be useful, because it summarizes the reasoning behind some of the choices in Arushi's pull request.

---

The short version of it is that faults are represented as a list of points on the surface and a 2D mesh perpendicular to that. The axis which aligns which the surface are called the sections. That is, the spaces between the coordinates you define are called sections. The axis going down are called segments, an each segment in each section can have a different length. The maximum_total_length variable is simply the maximum sum of all the segments. The thickness in the case of a fault is how far the fault extends outwards from that mesh/surface (half the thickness on both sides). This thickness can vary as well (both at the beginning and end of a segment), so this is just the maximum of all the segments in the grid. Each segment also has an angle at the start and at the end of a segment.

The tricky part is to determine how far the slab could possibly reach. Currently I have only tried to use very generic information like maximum length+thickness to set a maximum depth, because I know that there is no way that the slab can go any deeper than it's total length+thickness, no matter the angle. What Arushi is doing is using the same approach laterally. So take the min an max surface points and try to set a bound around it where the slab is simply too sort to reach. That is probably somewhere around being at a very low angle to the surface. In cartesian this is easy to do, but in spherical everything becomes more complicated. You would need to compute the bound of your box which encompasses all surface points in spherical coordinates, convert those to cartesian and extend the box in all directions with the maximum extend the slab could possibly go.

bangerth:
Do you have a picture somewhere in the documentation (or your paper) that illustrates all of these terms? I think that would really help me!

As for the "length": I think what you are talking about is the length of the *slab*, but the variable has the name *fault* in its name. That's what's confusing me. Maybe, again, a sketch or a picture would help :-)

In any case, what limits the depth? Is that done outside WorldBuilder?

MFraters:
> Do you have a picture somewhere in the documentation (or your paper) that illustrates all of these terms? I think that would really help me!
Does this help a bit? Currently the manual is mostly focused on communicating to users how to use the input files and not much on how the code internally works. 
drawing_distance_from_planes

> As for the "length": I think what you are talking about is the length of the slab, but the variable has the name fault in its name. That's what's confusing me. Maybe, again, a sketch or a picture would help :-)

Yes, the function was made for the slab, and then I realized that for faults you actually want to have the same type of shape, but instead of down down from the plane, having the plane in the center. 

> In any case, what limits the depth? Is that done outside WorldBuilder?

The location of the point defines by definition the distance from the bottom of the model to that point, and the depth defines how far that point is below the surface. Adding the two, gives you the total depth of the model. The reason behind this is that there can never be any discrepancy between the model definition in the Wold builder and for example Aspect. If you want to change your planet size or (more realistically) model depth, you do not need to change the world builder input file, just you aspect input file.

MFraters:
ah, and I realized yesterday night why the term fault length is so confusing. Because fault length is actually a term usually used to describe the length of the fault along the surface...

I am very much open to improved naming. I was already wanting to fix some of the naming in the fault plugins, because some of them even still have the name slab in them. I kinda wanted to wait for that until after #264 was merged to not change code close to that pull request too much while it was being made, but now I think I should have cleaned the names before to prevent some of the confusion. This would mostly consisted of the fault variable names changing variables from  `max distance slab top` to  something like `max distance fault center`. With what you mentioned, I am all ears for better names for the input and internal variables :)

bangerth:
On 7/7/21 9:28 AM, Menno Fraters wrote:
> ah, and I realized yesterday night why the term fault length is so confusing. 
> Because fault length is actually a term usually used to describe the length of 
> the fault along the surface...

Yes, that was my problem. @alarshi actually explained to me this morning what 
it means in the context of WB, so I finally understand.


> I am very much open to improved naming. I was already wanting to fix some of 
> the naming in the fault plugins, because some of them even still have the name 
> slab in them. I kinda wanted to wait for that until after #264 
> <https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/264> was merged 
> to not change code close to that pull request too much while it was being 
> made, but now I think I should have cleaned the names before to prevent some 
> of the confusion. This would mostly consisted of the fault variable names 
> changing variables from |max distance slab top| to something like |max 
> distance fault center|. With what you mentioned, I am all ears for better 
> names for the input and internal variables :)

I would say go ahead and we'll figure out all merge conflicts as they arise. 
You're at a point where a few more people are starting to poke around in the 
code and everyone is probably better off with more concise names -- not the 
least you yourself, given that you won't have to explain everything to 
everyone :-)

MFraters:
Oke, thanks for the advice. It is appreciated!

> You're at a point where a few more people are starting to poke around in the
code

Yea, I knew that the code was visible for everyone before, but now that I actually know people are really going through the code that actually feels scarier than I thought it would ;)

bangerth:
On 7/7/21 5:43 PM, Menno Fraters wrote:
> Yea, I knew that the code was visible for everyone before, but now that I 
> actually know people are really going through the code that actually feels 
> scarier than I thought it would ;)

Ha, but we're all friends and just want to understand and help!

MFraters:
And that is really appreciated!

PR #309, opened 2021-07-07
Rename a variable to clarify that it duplicates information of another variable.

bangerth:
Make it clear that one variable is the same as another variable, just expressed in a different coordinate system. Should I also rename `position` to `position_in_cartesian_coordinates`?

### DIFF
--- a/include/world_builder/features/continental_plate.h
+++ b/include/world_builder/features/continental_plate.h
@@ -85,7 +85,7 @@ namespace WorldBuilder
          * gravity and current temperature.
          */
         double temperature(const Point<3> &position,
-                           const  WorldBuilder::Utilities::NaturalCoordinate &position_in_natural_coordinates,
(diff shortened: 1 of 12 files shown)


MFraters:
This does indeed improve the readability a lot more than I expected :)

> Should I also rename position to position_in_cartesian_coordinates?

Yes, I think that would also make the code more readable. Thanks!

bangerth:
Let me do this as a follow-up, since it's not quite as automatic.

bangerth:
In other words, feel free to merge this one and I'll do the other variable in a separate PR. What name would you like? Is `position_in_cartesian_coordinates` accurate?

MFraters:
Sounds good to me. `position_in_cartesian_coordinates` is most consistent with `position_in_natural_coordinates` so that sounds good to me.

MFraters:
@alarshi This should fix the issue you had.

bangerth:
Related to #310.

PR #311, opened 2021-07-07
Fix bounding box 2d specialization

MFraters:
This introduces specific specializations for 2d and 3d and test them in the unit tests. What do you think @bangerth?


bangerth (commented):
Yes, this is exactly what was needed! Thanks for fixing my oversight!

Issue #310, opened 2021-07-07
Unclear notation


bangerth:
Here is another place where @alarshi and I got confused this morning regarding notation:
```
      // todo: explain
      const double starting_radius = natural_coordinate.get_depth_coordinate() + depth - starting_depth;
```
(This is in `Fault::composition()`, but I think it is also in a number of other places.) 

My confusion is rooted in the fact that *radius* is a length measure that is measured in *outward* direction, whereas *depth* is a length measure that is measured in *inward* direction. So one would expect that there would be different negative signs, but there aren't, and I'm confused as to what the variable corresponds to :-) 

In the end, this is in the following context:
```
    double
    Fault::composition(const Point<3> &position,
                       const  NaturalCoordinate &natural_coordinate,
                       const double depth,
                       const unsigned int composition_number,
                       double composition) const
    {
      // todo: explain
      const double starting_radius = natural_coordinate.get_depth_coordinate() + depth - starting_depth;
```
...which is called from here...
```
  double
  World::composition(const std::array<double,3> &point_,
                     const double depth,
                     const unsigned int composition_number) const
  {
    // We receive the cartesian points from the user.
    Point<3> point(point_,cartesian);

    WorldBuilder::Utilities::NaturalCoordinate natural_coordinate = WorldBuilder::Utilities::NaturalCoordinate(point,
                                                                    *(this->parameters.coordinate_system));
    double composition = 0;
    for (auto &&it : parameters.features)
      {
        composition = it->composition(point,natural_coordinate,depth,composition_number, composition);
```
...which is in turn called from ASPECT in....
```
    template <int dim>
    double
    WorldBuilder<dim>::
    initial_composition (const Point<dim> &position, const unsigned int n_comp) const
    {
      if (relevant_compositions[n_comp] == true)
        return this->get_world_builder().composition(Utilities::convert_point_to_array(position),
                                                     -this->get_geometry_model().height_above_reference_surface(position),
                                                     n_comp);
```
The way I see this chain is that the `position` and `natural_coordinates` express the same information, just in different coordinate systems. The `depth` variable is, by construction, again redundant, but this is not clear from the place where this is used. I *think* that one should be able to compute the depth from the given "natural" coordinates  if one were to knew a "reference depth" (or "radius"). But I don't know for sure; I would love to know, because that's what is necessary to compute bounding boxes.

MFraters:
[...]

PR #312, opened 2021-07-08
Rename more variables.

bangerth:
As discussed in #309 . I probably missed some places, but this covers a good number of them.


bangerth:
Related to #310.

MFraters (approved):
Thanks for doing the work of changing the names! I hope this will improve the readability :)

PR #313, opened 2021-07-08
Avoid using namespace to set config variables

gassmoeller:
I think this is the cleaner way to set these config variables. It usually does not matter much, except for unity builds, where the `using WorldBuilder` introduces all world builder names into the top namespace, which can lead to problems in downstream codes (see https://github.com/geodynamics/aspect/pull/4138). I hope nothing in world builder relies on this using statement, but it should not, because it is in a .cc file, not a header.

### DIFF
--- a/source/config.cc.in
+++ b/source/config.cc.in
@@ -19,17 +19,17 @@
 
 #include "world_builder/config.h"
 
-using namespace WorldBuilder;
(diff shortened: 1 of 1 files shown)


MFraters (approved):
Well, it seems to work both here and in the aspect pull request, so I will be happy to merge it :)

PR #314, opened 2021-07-08
Add some assertions to Point.

bangerth:
As requested in #301.


MFraters (approved):
Great, those are nice assert messages. Also thanks for fixing the variable names!

Once the old distance function has been removed in #304, we can also add an assert to prevent the current version of the dot product being used by a cartesian system.

MFraters:
oh, and the change in coverage is only for the assert messages, so that is fine.

PR #315, opened 2021-07-08
Remove an operator that is mathematically not well defined.

bangerth:
This is an operator of that would support the operation `3.141 / point`, but that is mathematically not well defined: We can divide a vector by a number, but we can't divide a number by a vector.

It turns out that this operator is (unsurprisingly) not actually used in WorldBuilder at the moment.


MFraters (approved):
Well, it is good to know the world builder is not using a mathematically operation which is not well defined, even though it defines it.

PR #316, opened 2021-07-08
WIP: remove non contiuous option from distance point from curved planes [v2]

MFraters:
This is a replacement for #304, because I would like to cut that pull request up in multiple steps. It completely removes the option to interpolate coordinates in the feature interface. I don't think these kind of options should be available on that level, because already some options do not make sense for all features (continuous for continental plates). I have not removed the enum for the interpolation types, since I might want to make this kind of interpolation available again for the area features (oceanic/contineteal/... plates). For slabs and faults the only option left will be the continuous monotone spline. 

Todo:
- [ ] find what to do with the current top level interpolation type. I am thinking of removing it
- [ ] remove one dimensional coordinates from spline.

I am still doubting about these two todo's since they also have consequences for area features. For the slab and fault I have convinced myself that the continuous function is always the best. 

Also, I have not done my best to do optimalization here, but I have changed the `distance_point_from_curved_planes` such that there is already room for future optimizations.

@bangerth feel free to take a look at this. It should be much more readable than #304.

### DIFF
--- a/include/world_builder/utilities.h
+++ b/include/world_builder/utilities.h
@@ -194,7 +194,7 @@ namespace WorldBuilder
         /**
          * x coordinates of points
          */
-        std::vector<double> m_x;
+        size_t mx_size_min;
 
         /**
          * interpolation parameters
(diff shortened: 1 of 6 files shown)


MFraters:
I need to note for myself that in this pull request, I can get rid of m_x in the interpolation scheme, because no additional parts are needed. So, x can just be `0,1,2,3,...`, which means that it can be assumed.

PR #320, opened 2021-07-20
add depth of closest point on reference surface to struct.

MFraters:
Adding a variable in `PointDistanceFromCurvedPlanes` storing the depth of the closets point on the reference surface from the `distance_point_from_curved_planes` function. 

This functionality is needed for a better slab temperature being developed by @mibillen. 

I have tested it locally (see pictures), but will need to add tests to the unit tester.

Screenshot from 2021-07-19 17-46-42
Screenshot from 2021-07-19 17-46-16


MFraters:
A few tests have been added to the unit tester. Some of them are checked, others are mostly regression tests.

MFraters:
A few more tests and now with a better color scheme (vik). It all looks correct to me. 

I would like to make these into official tests, but I would first net to come up with a debug/develop plugin set structure for this (see #321). 

Screenshot from 2021-07-20 11-18-52
Screenshot from 2021-07-20 11-18-17
Screenshot from 2021-07-20 11-16-04
Screenshot from 2021-07-20 11-11-07

PR #322, opened 2021-07-28
Remove an unused variable.

bangerth:


MFraters (approved):
Sorry for the late reply, I have been on holiday and am now catching up. 

Good catch! This is a left over from the plugin this was based on.

PR #326, opened 2021-08-17
Avoid the use of intermediate storage arrays in computing bounding boxes.

bangerth:
Follow-up to #264.

### DIFF
--- a/source/features/fault.cc
+++ b/source/features/fault.cc
@@ -331,27 +331,22 @@ namespace WorldBuilder
       // Find minimal and maximal coordinates. Do this by finding the
       // leftmost/rightmost point with regard to either the [0] or [1]
       // coordinate, and then takes its [0] or [1] element.
-      min_along_x = (*std::min_element(coordinates.begin(), coordinates.end(),
-                                       [](auto p1, auto p2)
+      auto compare_x_coordinate = [](auto p1, auto p2)
       {
         return p1[0]<p2[0];
-      })) [0];
-      max_along_x = (*std::max_element(coordinates.begin(), coordinates.end(),
-                                       [](auto p1, auto p2)
-      {
-        return p1[0]<p2[0];
-      })) [0];
+      };
 
-      min_along_y = (*std::min_element(coordinates.begin(), coordinates.end(),
-                                       [](auto p1, auto p2)
+      min_along_x = (*std::min_element(coordinates.begin(), coordinates.end(), compare_x_coordinate)) [0];
+      max_along_x = (*std::max_element(coordinates.begin(), coordinates.end(), compare_x_coordinate)) [0];
+
+
+      auto compare_y_coordinate = [](auto p1, auto p2)
       {
         return p1[1]<p2[1];
-      })) [1];
-      max_along_y = (*std::max_element(coordinates.begin(), coordinates.end(),
-                                       [](auto p1, auto p2)
-      {
-        return p1[1]<p2[1];
-      })) [1];
+      };
+
+      min_along_y = (*std::min_element(coordinates.begin(), coordinates.end(), compare_y_coordinate)) [1];
+      max_along_y = (*std::max_element(coordinates.begin(), coordinates.end(), compare_y_coordinate)) [1];
 
       min_lat_cos_inv = 1. / std::cos(min_along_y);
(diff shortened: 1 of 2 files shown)


bangerth:
@alarshi FYI.

alarshi:
@bangerth : Thank you for sharing the modification, I did not know it could be computed this way.

MFraters:
Thanks for the pull request. Could you elaborate a bit on what the advantage of using lambda functions here is? It is turning four short one line functions into four 5 line function. Is this faster for you computations?

bangerth:
The idea is this: You already have the `coordinates` array, and you'd like to find the smallest `x` value. The current code does this by copying all `x` coordinates into a separate array and then finding the smallest value via `min_element`. What my approach does is to just work on the original `coordinates` array and re-defining what "smallest" means: That's because there is no way to say "this point is smaller than that point" via `operator<`, but I can define a (lambda) function that compares two points by looking at their `x` coordinates. So the lambda functions are used to pick out which is the left-/right-/bottom-/top-most point, and at the very end I then only need to query the `x` or `y` components of these points.

The code may be slightly more difficult to read, but it avoids the need to copy the data into two arrays.

(It's ok if you don't like the approach and don't want to merge it. It's just something I saw and thought could be made more efficient. But I'm not married to the patch :-) )

MFraters:
hmm, I can see where you are coming from and that should definitely be more efficient, although I am not seeing much in the benchmarks. 

The more I think about it, I do like the approach of not having to copy stuff around needlessly. I am wondering though whether it can be simplified into for example this:
```
      auto compare_along_x = [](auto p1, auto p2)
      {
        return p1[0]<p2[0];
      };

      auto compare_along_y = [](auto p1, auto p2)
      {
        return p1[1]<p2[1];
      };

      min_along_x = (*std::min_element(coordinates.begin(), coordinates.end(),compare_along_x))[0];
      max_along_x = (*std::max_element(coordinates.begin(), coordinates.end(),compare_along_x))[0];
      min_along_y = (*std::min_element(coordinates.begin(), coordinates.end(),compare_along_y))[1];
      max_along_y = (*std::max_element(coordinates.begin(), coordinates.end(),compare_along_y))[1];
```
This compiles and passes the tests on my computer, and looks easier to read to me. I think it would be even better if the `compare_along_x` and `compare_along_y` functions could be merged, although we now have the slightly weird `dimension_index` variable which needs to be created before the labmda function and change halfway:
```
      unsigned int dimension_index = 0;
      auto compare_along_dim = [dimension_index](auto p1, auto p2)
      {
        return p1[dimension_index]<p2[dimension_index];
      };

      min_along_x = (*std::min_element(coordinates.begin(), coordinates.end(),compare_along_dim))[dimension_index];
      max_along_x = (*std::max_element(coordinates.begin(), coordinates.end(),compare_along_dim))[dimension_index];
      
      dimension_index = 1;
      min_along_y = (*std::min_element(coordinates.begin(), coordinates.end(),compare_along_dim))[dimension_index];
      max_along_y = (*std::max_element(coordinates.begin(), coordinates.end(),compare_along_dim))[dimension_index];
```
This also compiles and passes the tests for me. What is your opinion on these options?

bangerth:
You're totally right with making this more readable. I've adapted the first idea. I don't think the second one is a benefit because you're changing what the function does based on the `dimension_index` variable.

MFraters:
Yes, that looks more readable to me, thanks for making the changes! Could you also apply the same changes to the subducting plate file?

As for the coverage testing, the coverage has not been changed, but for some reason it doesn't recognize `min_along_x = (*std::min_element(coordinates.begin(), coordinates.end(),` and `compare_x_coordinate)) [0];` as being the same statement and only attributing coverage to one of them. Putting them on one line may or may not help, although this is clearly a gcov issue, so this is not blocking a merge. Besides the coverage issue, I personally think it would be more readable if those lines would become one liners instead of being split, since it will be easier to see that the only difference is min element and max element, and I don't think the one liner would be too long. Is there a specific reason you put it on two lines?

bangerth:
Done as suggested!

MFraters (approved):
Thanks for the changes and also the small fixes you made along the way :)

This is ready to merge now!

PR #327, opened 2021-08-17
Make code easier to read.

bangerth:
The current code does
```
  Object x;
  sometype &y = x.internal_representation();
  y.a = ...;
  y.b = ...;
  return x;
```
This is hard to read, primarily because to the reader, the association between `x` and `y` is lost. But it's easy to avoid this.

Follow-up to #264. @alarshi FYI.

### DIFF
--- a/source/features/fault.cc
+++ b/source/features/fault.cc
@@ -355,30 +355,29 @@ namespace WorldBuilder
     Fault::get_bounding_box (const NaturalCoordinate &position_in_natural_coordinates,
                              const double depth) const
     {
-      BoundingBox<2> surface_bounding_box;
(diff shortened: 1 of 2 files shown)


MFraters:
I like the simplification. Although it looks the correct to me (same code, just some reorganization), for some reason it fails the debug tests, complaining about the following for the spherical tests: 
```what():  Assert `boundary_points.first[i] <= boundary_points.second[i]` failed in /home/runner/work/WorldBuilder/WorldBuilder/include/world_builder/bounding_box.h at line 264: Bounding Box can't be created: the points' order should be bottom left, top right!``` 
for two spherical tests: `24 - spherical_fault_interpolation_simple_none` and `25 - spherical_slab_interpolation_simple_CMS`. 

This could either indicate that there is some small mistake I am missing or that there may be a bug in the original code. Could you have another quick look at it to see if you can find it?

PR #334, opened 2021-09-24
Add c++14 flags to tester.

MFraters:
This should help with some tester issues @rfildes found on her mac. Does not address all issues.


MFraters:
The tests are fine (one is not reporting to github somehow, but the results look fine) so there doesn't seem to be any harm in merging this.

PR #336, opened 2021-09-27
Add macos-11 to tester.

MFraters:
Currently the the tester macos-latest is on 10. Here I explicitly add a tester for osx 11.


MFraters:
Needed to add some library paths to make osx 11 work. This should fix the last issue related to #334

PR #344, opened 2021-10-22
add temperature plate model constant age to oceanic plate models

lhy11009:
I add a "plate model constant age" plugin under "oceanic plate". This takes a constant age of the plate and assign a temperature field using the plate model.


lhy11009:
Hi
@MFraters, I am not sure what's the problem here. It complains that there is no test point that is out of the range defined for the feature, is that correct? While the "max depth" defined in the test wb file is 250e3 but there is query points at 300e3 (dat) file, then why is this the case? Can you help to explain it a bit, thankss?

MFraters:
Great, thanks for the pull request and making the changes!

PR #356, opened 2021-10-29
Smooth composition for transition of faults.

alarshi:
This PR adds a composition model in faults so that the composition changes value smoothly from 1 to zero.


MFraters:
Cool, thanks for sharing this! I think this will be useful for many more users :) To start off, I have a few comments:

1. In the world builder termololgy, compositions are the labels (integers). The actual value of the compositions are called fractions (which are doubles), since they store what fraction of the composition is present. I think therefore `center composition` and `side composition` should be renamed to  `center fraction` and `side fraction`.
2. Currently you are not using `side compositions`, so we will need to decide whether to use it or remove it for now (we can always add it later.
3. I think there is a lot of value to generalizing the code a little bit. For example, the uniform plugin allows for multiple compositions and fractions. Do you think that is something which would make sense for this plugin?
4. Could you add a test? If you need help with how to add tests for the world builder, let me know.
5. If you have some images of the composition distribution in a fault, feel free to post it here in the pull request!

alarshi:
Hi Menno, thank you for the review! :) I am sorry for the delay in addressing your comments.

> 1 ) The actual value of the compositions are called fractions (which are doubles), since they store what fraction of the composition is present. I think therefore center composition and side composition should be renamed to center fraction and side fraction.

I understand, I modified the terminology for ```side_fractions``` and ```center_fractions```.

> 2) Currently you are not using side compositions, so we will need to decide whether to use it or remove it for now (we can always add it later.

Yes, we were not using the value because the default worked for us. I have added ```side_fraction``` in calculation of composition. So, depending on the user-specified value, the composition fraction can go from ```center_fraction``` to ```side_fraction``` over a distance. But you are right, it is probably not very useful for practical purposes.

>3) I think there is a lot of value to generalizing the code a little bit. For example, the uniform plugin allows for multiple compositions and fractions. Do you think that is something which would make sense for this plugin?

Yes, that makes sense. I modified the structure in the recent commit so we can have multiple compositions and fractions.

>4) Could you add a test? If you need help with how to add tests for the world builder, let me know.

Yes, I would be happy to write a test! I know how to write a test in Aspect but reading the WB manual, it looks like I might need some help. :-/ 

5> Here is one of the images that demonstrate the differences between using the ```uniform``` and ```smooth``` compositions in faults. 
image_faults
Fault composition using the uniform (left) and the smooth (right) fault models. 

Here is one of the zoomed images of faults showing the points:
image (1)

Let me know if it makes more sense as a fault model to have in the WB. It was useful for us because our viscosities were dependent on the fault composition, so a smooth transition also ensured smoothness in the viscosity.

MFraters:
For of all, I am sorry for letting this go unanswered for so long, it has been a bit of a hectic period :(

To start with your last question, yes, I do think this will be useful to other users as well, so please do not be discouraged by my very late reply.

> Yes, I would be happy to write a test! I know how to write a test in Aspect but reading the WB manual, it looks like I might need some help. :-/

That is definitely something which needs to be expanded on in the manual :) 

In short there are several different kind of tests available for the world builder. They can all be found in the tests folder. 

1. Unit tests: these are the same as in ASPECT. There are mostly meant to test very specific functions.
2. app tests: The world builder App is a terminal application which takes a world builder file and an ascii data file with coordinates (and some extra info) and return an ascii file with the input appended with temperature and other output per line. This is the kind of test I would suggest you make. The easiest way to make this is copy an existing test and give it a new name. Then change the .wb file to a relatively simple case which uses this plugin. Next visualize it either through ASPECT or through the world builder visualizer and pick some coordinates in and outside the fault and add those to the .dat file. When you run the app check whether the output is correct and overwrite the screen-output.log in the correct folder (like you do with an ASPECT test). 
3. Visualization test. These work very similar to the app test but instead call the world builder visualizer and check the vtu ouput against a reference vtu output. At the moment the goal of this is mostly to test the visualizer itself.
4. test for several language interfaces: tests whether the C, CPP (MPI), Fortran and Python still function.

I hope this helps a bit. Like I said, I think the best test to add for you case will probably be 2: the app test. If you have any questions about this, or want to discuss this over zoom, let me know.

alarshi:
> For of all, I am sorry for letting this go unanswered for so long, it has been a bit of a hectic period :(

No worries at all. I know with the AGU and the end of semester, it would have been extremely busy time period.

> I hope this helps a bit. Like I said, I think the best test to add for you case will probably be 2: the app test. If you have any questions about this, or want to discuss this over zoom, let me know.

Thanks for explaining how to write the tests! I will give it a go and if I have any troubles, I will schedule a meeting with you.

MFraters:
@alarshi What is the status of this pull request? Would be great to get this merged :)

alarshi:
@MFraters : Hi Menno, I am sorry but this semester has been a bit busy with teaching and research and I did not address your suggestions on this PR. I still have it on my mind and the semester will be done in a week so if you don't mind keeping this open until then, I will definitely get back to it!

MFraters (approved):
Thank contributing this and finishing the pull request!

PR #359, opened 2021-10-29
fix macos-latest tester.

MFraters:
fix macos-latest tester and maybe update xcode.


MFraters:
This seem to fix the issue with the the tester, for now...

Issue #353, opened 2021-10-28
Slab tip temperature anomaly below the subducting plate for the "mass conserving" model


lhy11009:
I discussed with @mibillen this morning the discontinuity of the slab temperature below the subducting tip.
Initially, if we look at the subducting plate (right side), the temperature below the slab tip has to increase from the temperature from a plate model to the mantle temperature.
![um_temperature000005](https://user-images.githubusercontent.com/28665365/139344438-9e826f69-7c87-471f-9511-d8b12011279f.png)
Later on in the model run, it develops as a little dripping body:
![um_temperature000013](https://user-images.githubusercontent.com/28665365/139344571-7c1d23e6-b660-4b86-8364-394bf38da192.png)
This would mass up the velocity a bit:
![um_viscosity_snap000013](https://user-images.githubusercontent.com/28665365/139344618-cfffe62a-fe53-428f-b5e8-417427bf11c3.png)
This might be an issue since I assigned a very short slab there. I would like to look into if we can fix that. Maybe, add another boundary layer below the slab tip?
Thne I changed the inputs in the .wb file to have two segments:

{ 
        "model":"subducting plate", "name":"Slab",
        "coordinates":[[4e6,-5e6],[4e6,5e6]],
        "dip point":[10e6,0],
        "segments":[
        {
            "length":314.16e3, "thickness":[300e3], "top truncation":[-100e3], "angle":[0,90],
	        "composition models":[
	        {"model":"uniform", "compositions":[0], "max distance slab top":7.5e3},
	        {"model":"uniform", "compositions":[1], "min distance slab top":7.5e3, "max distance slab top":35.2e3}
            ]
        },
        {"length":100e3, "thickness":[300e3], "top truncation":[-100e3], "angle":[90,90]}
        ],
        "temperature models":
        [
        {"model":"mass conserving", "density":3300, "thermal conductivity":3.3,"adiabatic heating":true, 
         "plate velocity":0.05,"ridge coordinates":[[0,-5e6],[0,5e6]], 
         "coupling depth":200e3, "shallow dip":60.0, "taper distance":100e3, 
         "min distance slab top":-15e3, "max distance slab top":300e3}
        ]  
    }

![Selection_999(345)](https://user-images.githubusercontent.com/28665365/139485373-419a096a-e68c-4705-9af6-318be0d8ef42.png)

lhy11009:
### Then I changed the inputs in the .wb file to have two segments:

{ 
[...]

PR #360, opened 2021-10-30
taper slab age at tip 

mibillen:
so underside of slab thins as the min temperature increases to the background temperature.
This should address issue #353.
However, in looking at this there is still an issue with the "background temperature" used for the top side of the slab where is needs to blend into the overriding plate temperature. I will look at this next week.


MFraters (approved):
Looks good to me. Could you indent the code and update the test results?

mibillen:
I fixed the indenting and I also fixed the issue with the temperature on the overriding plate - there were temperatures that were less than the surface temperature  in the region where the slab has a very shallow dip. This is because even though there is very little negative heat content going into the top of the slab, the overriding plate temperature is very close to or even less than the minimum temperature of the slab. Therefore, if the overriding plate temperature is less than the minimum slab temperature, then we don't change the temperature there.  There is still small step in the temperature contours (for the same reason... small deltaT requires a large distance to spread out heat content, even when its not that much). I don't see a way around this. To minimize the size of the offset, I recommend making the "min distance top slab" = -100e3 for this shallow segment. 
This is not an issue when there is no overriding plate (see image). 
Screen Shot 2021-10-30 at 5 07 18 PM
Here's my final version with the overriding plate
Screen Shot 2021-10-30 at 5 10 39 PM
e

MFraters:
The negative temperature in the overriding plate is actually something I had noticed but didn't have time to investigate yet. Good to hear that it is now fixed!

The pictures and it also seem to resolve #353. Can you update the test results?

I also think that these fixes are notable enough to add a line to the changelog.md file, so feel free to add a line there in the fixed heading.

mibillen:
I changed the test for cartesian to include an overriding plate and a dipping slab so that I would end up with points above the slab that are colder than the minimum temperature. This meant that I also need to change the location of the test points in the grid file. This was needed to insure all parts of the code are covered. Here are pics of the new test case.
Screen Shot 2021-11-03 at 3 29 26 PM
d
Screen Shot 2021-11-03 at 3 29 41 PM
.

MFraters:
Thanks for fixing the coverage as well!

Issue #331, opened 2021-09-14
Define model geometry extend in or to the world builder.


MFraters:
This came up in a discussion with @bangerth during the aspect biweekly meeting and is related to https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/318, and may be of interest to @alarshi. The post has become a little bit longer than I hoped, but I think it is a good summary of the issue and possible solutions.

The main issue is that it is currently very hard to optimize for especially spherical domains, because the world builder is designed to not know anything about the domain, but fully rely on the depth and height information provided at each request. The advantage of this approach is that there is no room for error by the user when building a model, because the code calling the world builder defines the exact world in the call. For example in ASPECT the user defines the information about the world in the ASPECT input file and when aspect requests the temperature at a certain height and depth, the radius of the earth in a spherical geometry is automatically defined based on that information.

This prevents mistakes and provides a lot of flexibility, because you can change your aspect model domain without having to worry about the world builder file and even change the size of the world by modifying the the values you provide to the get function. The downside is that the GWB only knows the size of the planet at the time it is requested. This also means that many optimizations, especially related to the slabs (which are expensive to compute) are simply not possible. 

A different issue, which was mentioned by @tjhei, is that this also means that a GWB file does not uniquely define a model. It only defines a unique model in combination with for example an ASPECT input file or a world builder visualizer input file. This may not be what users expect.

Here are some of the proposed solutions:

1. Provide the radius of the earth or model box depth/height for a spherical or Cartesian geometries respectively as a **special optimization input parameter** in the GWB input file. This would be be easy to implement, but it would hand of the responsibility to the user, creating unexpected results for users who are just reusing files they got from others. 
2. Provide the radius of the earth or model box depth/height for a spherical or Cartesian geometries respectively as a **normal input parameter** in the GWB input file in the coordinate system section. This would be be easy to implement and would change the world builder file into a complete description of the initial conditions, but it would allow for incompatibilities between the GWB and ASPECT input files. In the current structure that would mean adding an assertThrow (release mode) in every query to prevent this. I think this would be error prone and annoying to users (see option 5). 
[...]

PR #363, opened 2021-11-03
Replace INFINITY by std::numeric_limits.

bangerth:


MFraters (approved):
As discussed in #331, this a better way of setting infinity. Thanks!

PR #367, opened 2021-11-08
Add max_model_depth variable to cartesian and spherical geometries 

MFraters:
This is a first step based on the discussion in both #331 and #345. This only add max_model_depth variable to cartesian and spherical geometries with a get function and a radius input parameter for the spherical geometry model to define the `max_model_depth`. 

Merging this will allow to start writing code which uses this internally and within ASPECT.

### DIFF
--- a/source/coordinate_systems/cartesian.cc
+++ b/source/coordinate_systems/cartesian.cc
@@ -83,6 +83,13 @@ namespace WorldBuilder
       return point_at_depth.norm();
     }
 
+
+    double
+    Cartesian::max_model_depth() const
+    {
+      return std::numeric_limits<double>::infinity();
+    }
+
     /**
      * Register plugin
      */
--- a/source/parameters.cc
+++ b/source/parameters.cc
@@ -1383,7 +1383,51 @@ namespace WorldBuilder
           }
         else if (type == "object")
           {
-            collapse += "/properties";
+            // normal objects just have properties, but some have oneOf
+            if (Pointer((base_path + "/oneOf").c_str()).Get(declarations) != nullptr)
+              {
+                // it has a structure with oneOf. Find out which of the entries is needed.
+                // This means we have to take a sneak peak to figure out how to get to the
+                // next value.
+                size_t size = Pointer((base_path + "/oneOf").c_str()).Get(declarations)->Size();
+#ifdef debug
+                bool found = false;
+#endif
+                size_t index = 0;
+                for (; index < size; ++index)
+                  {
+                    std::string declarations_string = Pointer((base_path + "/oneOf/" + std::to_string(index)
+                                                               + "/properties/model/enum/0").c_str()).Get(declarations)->GetString();
+
+                    // we need to get the json path relevant for the current declaration string
+                    // we are interested in, which requires an offset of 2.
+                    WBAssert(Pointer((get_full_json_path(i+2) + "/model").c_str()).Get(parameters) != nullptr, "Could not find model in: " << get_full_json_path(i+2) + "/model");
+                    std::string parameters_string = Pointer((get_full_json_path(i+2) + "/model").c_str()).Get(parameters)->GetString();
+
+                    // currently in our case these are always objects, so go directly to find the option we need.
+                    if (declarations_string == parameters_string)
+                      {
+                        // found it for index i;
+#ifdef debug
+                        found = true;
+#endif
+                        break;
+                      }
+                  }
+#ifdef debug
+                WBAssert(found == true,
+                         "Internal error: This is an array with several possible values, "
+                         "but could not find the correct value " << collapse + "/" + path[i] + "/items/oneOf");
+#endif
+                collapse += "/" + path[i] + "/oneOf/" + std::to_string(index) + "/properties";
+                // add one to i, to skip the array
+                ++i;
+
+              }
+            else
+              {
+                collapse += "/properties";
+              }
           }
         else
           {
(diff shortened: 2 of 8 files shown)


source/coordinate_systems/cartesian.cc
```diff
@@ -83,6 +83,13 @@ namespace WorldBuilder
       return point_at_depth.norm();
     }
 
+
+    double
+    Cartesian::max_model_depth() const
+    {
+      return std::numeric_limits<double>::infinity();
+    }
```
bangerth, line 91: Out of curiosity, why infinity? Does the model know anything about its lateral extent?

I'm asking because you've in essence described a bounded domain (a sphere of known radius) in the spherical case, but this corresponds to an unbounded domain.

source/parameters.cc
```diff
@@ -1383,7 +1383,51 @@ namespace WorldBuilder
...
+            // normal objects just have properties, but some have oneOf
+            if (Pointer((base_path + "/oneOf").c_str()).Get(declarations) != nullptr)
+              {
+                // it has a structure with oneOf. Find out which of the entries is needed.
+                // This means we have to take a sneak peak to figure out how to get to the
+                // next value.
+                size_t size = Pointer((base_path + "/oneOf").c_str()).Get(declarations)->Size();
+#ifdef debug
+                bool found = false;
+#endif
```
bangerth, line 1395: This has nothing to do with the patch itself, but just as a comment: I assume that you are putting the `#ifdef` around this statement because you would otherwise get a warning about the `found` variable only ever being written to, but never read, in release mode? If so, you can avoid the warning if you put
```
  (void)found;
```
in front of the `WBAssert`. This is interpreted as a "read" of the variable, but you are just discarding the value so read.

tests/unit_tests/unit_test_world_builder.cc
```diff
@@ -3823,9 +3823,12 @@ TEST_CASE("WorldBuilder Parameters")
 {
   // First test a world builder file with a cross section defined
   std::string file = WorldBuilder::Data::WORLD_BUILDER_SOURCE_DIR + "/tests/data/type_data.json";
-  std::string file_name = WorldBuilder::Data::WORLD_BUILDER_SOURCE_DIR + "/tests/data/subducting_plate_different_angles_spherical.wb";
+  std::string file_name = WorldBuilder::Data::WORLD_BUILDER_SOURCE_DIR + "/tests/data/subducting_plate_different_angles_cartesian.wb";
```
bangerth, line 3826: Aren't you abandoning the old test here?

source/coordinate_systems/cartesian.cc
MFraters: My idea was that to fully define a position in a sphere based on depth, need a radius. For a Cartesian models that is not needed. You can go to infinity in any direction without any problem. 

I guess I could make this an option anyway within the world builder for optimization reasons, or if you really want to cut it off at some depth (which you can already do). Is that what you mean?

source/parameters.cc
MFraters: interesting, I will try that!

tests/unit_tests/unit_test_world_builder.cc
MFraters: I noticed the only thing happening with `file_name` here is `WorldBuilder::World world(file_name);` and then `Parameters prm(world);`. Then the prm is actually initialized with a different file. These first two steps on this exact file are already being tested in line 7099-7100, so there is no reason to repeat that here. And there was no place in the unit tester which did this for the Cartesian case, so I think it is a nice expansion of the testing without losing anything. 

I could have made a new test as well, but I guess I was a bit lazy ;)

MFraters (commented):
Thanks for the review!

source/parameters.cc
MFraters: on my laptop those if statement do not seem to matter, but probably for one of the tester configurations it does. I will look into it in a different pull request.

source/coordinate_systems/cartesian.cc
bangerth: Right, I understand the need for spherical objects. I was mostly wondering how you conceptually think of Cartesian worlds: do they have lateral extents that are defined somewhere, or are they infinite in the lateral directions? If the former, then it's worthwhile also specifying a maximal depth; in the latter case, there is no need for that.

MFraters: There are no lateral extends defined in Cartesian coordinates, the domain is infinite.

MFraters:
@bangerth I think that all your comments have been addressed. I will wait with merging this for few hours in case you have more comments.

bangerth:
I won't get to it before tomorrow afternoon, but feel free to go ahead.

source/coordinate_systems/cartesian.cc
bangerth: I see. In that case, this makes sense.

MFraters:
I am happy to wait if you want to have another look at it :)

bangerth (commented):
I have no further comments about the patch, so go ahead with merging! I guess the next step will be to propagate this information to the feature plugins?

MFraters:
Great, thanks! With this pull request the information should now be available anywhere in the world builder, so it will be both using it and making the changes to the interface which now have become possible.

bangerth:
Sounds good. Feel free to tag me in any follow-up patches for which you are looking for input!

PR #370, opened 2021-11-10
Add gravity model plugin system

MFraters:
This is a proposed solution to the idea of removing the gravity from the temperature query interface as proposed in #331. This pull request implement a gravity model system which defaults to a uniform gravity of 9.81. Other possible plugins would be a linear gravity plugin (like what aspect has) and a prem gravity model. More complicated gravity functions would also be possible by loading in data sets or prescribing functions.

If this is the desired direction then it would also be easy to add a new external interface where other applications can query the gravity.


MFraters:
This pull request not only adds a gravity plugin system, but also replaces the `gravity_norm` in the external temperature interface with this internal gravity plugin. This will allow for the removal of this parameter from the external temperature interface in a follow up pull request.

PR #372, opened 2021-11-11
Improve latex declarations readability in the manual.

MFraters:
The World Builder file parameter reference section in the manual is very hard to read because the nested structure it is trying to represent is very deep (up to 14 levels) and very big/long. This is partly a limitation with latex and moving to a html based manual or a xml file interface like aspect has may be best. But I think I found some setting which do improve the situation. 

1. Add the depth of the current header in front of the header: `(4) /this/has/four/levels/`
2. Indent the information in each section/header with 1em per level. 

This is the results, which I think looks easier to read to me. I will leave the pull request open for a few days in case there are comments or objections. 

example_new_manual_parameters


MFraters:
I got some input and it was positive, so I am going to merge this.

PR #377, opened 2022-02-06
Save a few cycles when outputting only a single char.

bangerth:
Inspired by https://github.com/dealii/dealii/pull/13339.

/rebuild

### DIFF
--- a/app/main.cc
+++ b/app/main.cc
@@ -166,14 +166,14 @@ int main(int argc, char **argv)
             std::cout << "# x z d g T ";
 
             for (unsigned int c = 0; c < compositions; ++c)
-              std::cout << "c" << c << " ";
+              std::cout << 'c' << c << ' ';
 
             for (unsigned int gc = 0; gc < grain_compositions; ++gc)
               for (size_t g = 0; g < number_of_grains; g++)
-                std::cout << "gs" << gc << "-" << g << " " // gs = grain size, gm = grain rotation matrix
-                          << "gm" << gc << "-" << g << "[0:0] " << "gm" << gc << "-" << g << "[0:1] " << "gm" << gc << "-" << g << "[0:2] "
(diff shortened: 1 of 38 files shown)


MFraters:
Interesting, thanks for bringing this also to the world builder. I am kind of wondering whether this actually makes a difference, since I kind of thought this kind of thing would be optimized by the compiler. Anyway, I would be happy to merge this, but please exclude the external libraries vtu11 and catch2 (the last one will also help with fixing the tester).

bangerth:
I've undone the changes to the external libraries.

As for compiler optimizations: C++ has a hard time with these sorts of things because the standard library is by and large implemented in header files that one can replace at will -- in other words, you are free to provide your own overloads to
```
  ostream & ostream::operator<< (const char *)     // for "abc"
```
and
```
  ostream & ostream::operator<< (const char)       // for 'a'
```
As a consequence, the compiler can't replace a call to one by a call to another. In the end, putting a single character into a stream requires just a single-char copy, whereas putting `"a"` into a stream requires calling `strlen` to determine the length of the string, followed by a call to `memcpy` (which will be inefficient since it contains a loop that runs only once).

I would be greatly surprised if one can measure the effect of this patch in seconds or milliseconds, but it was low-hanging fruit to fix given that I had the right regex already sitting around.

MFraters (approved):
Great, thanks for the changes. 

Also thanks for the explanation. It doesn't seem to make any significant difference in the benchmarks, but based on your explanation, I think it is just good practice. 

Could you post (or link to) the regex you used in this pull request for future reference?

bangerth:
I *think* it was something like `perl -pi -e "s/<< \"([^\'])\"/<< \'\1\'/g;" `find . | egrep '\.(h|cc)'``, but that required a bit of manual massaging because there are places (at least in deal.II and ASPECT) where we had strings of the form `"\"abc"`. I believe something of the sort `perl -pi -e "s/<< \"([^\'\\])\"/<< \'\1\'/g;" `find . | egrep '\.(h|cc)'`` might work, but I have not tested that.

MFraters:
Thanks!

Issue #379, opened 2022-02-06
[Tracking issue] New GWB manual


MFraters:
This issue is to track the progress on the new manual. For info on why this is done, see bottom of the post. 

The new manual is made with MyST (a hybrid between markdown and reStrcuturedText), build with sphinx, using the book theme. The theme seems to work well for different screen sizes. I can also create a latex pdf manual from the source which looks quite good as well. 

For the structure, the idea is to have a lot of small pages with a clear focus. That should improve readability and looking up specific information. 

The style and outline of the new manual are getting there, but require a bit more work before a first pull request can be made. Below is my plan:
- [x] setup layout and code structure for the new manual
- [x] Setup content structure will all pages in place with a one or two sentence todo-box explaining what information the page should contain. 
- [x] Make a pull request of that. This pull request will not be reviewed on content (except for large mistakes), only on whether this is the correct approach and on functionality. I want to have merged quickly so that I can set up a read-the-docs page and connect the github edit button correctly. That should allow for a much more interactive and iterative process. See #381 
- [x] Start the process of getting feedback on the content structure: is there information missing, are there pages missing, could pages be removed or reordered, etc. This is the point at which I will start involving all collaborators, and anyone else who is interested. 
- [x] Process feedback on the content structure. #387 
- [ ] Start filling in the content to the pages. 
- [x] Add citations #391 
- [x] Let world builder create markdown files instead/in addition to latex files for the parameters #390 
- [ ] Fix the chat link issue

I want to stress that feedback on the process or content is welcome at any point. 

![new_manual_v01](https://user-images.githubusercontent.com/7631629/152670526-184947f2-6e25-40c2-a70a-9be960bc8789.png)
 
Reason for the new manual: The current manual was written for initial release. Since then, it has been improved upon and added to and the code has seen increasing adoption. From user feedback and to aid in learning how to build complex models, I plan to add a new step-by-step new user tutorial, more examples for 2D and 3D models and additional documentation on parameterization of model features and properties. In targeting code contributions, the manual will also be updated to better describe the code structure.

@gassmoeller

MFraters:
An outline of the new manual is now online and can be found on: https://gwb.readthedocs.io

gassmoeller:
This looks incredibly fancy :-) ... I like it.

bangerth:
I agree, that's a nice web page!

MFraters:
Thanks both! I am also happy with the direction it is going in :)

PR #381, opened 2022-02-07
New sphinx based manual

MFraters:
This pull request is addressing point 2 and 3 in issue #379. Besides the infrastructure to build html and latex sphinx through cmake, it also include the infrastructure to start building doxygen through cmake. One of the advantages is that it will automatically update the version. It will also be useful if we ever want to switch to putting everything in sphinx (which I interested in, but not convinced of of yet).

Below are some examples of what this will look like.

Home page:
front_page

Install guide page with synchronized tabs for different operating systems:
istallation_guide_page

Unit testing page with cross-reference and showing the github button allowing to directly edit the page through github:
unit_test_page_with_github_button


MFraters:
The current state is I think sufficient for the purpose of stating to ask feedback on the new manual layout and content outline. Although quite some improvements on the structure and infrastructure remain to be done, I am happy with the outline of the manual. To be able to do some final tests (for example the github connection), this needs to be merged.

This pull request is also adding documentation testing, which was much easier than expected and will help a lot with making sure the manual input files stay up to date and functional.

Issue #386, opened 2022-02-09
Feedback on outline of the new manual


MFraters:
This issue is to receive feedback on the outline of the new manual, which can be found at [https://gwb.readthedocs.io](https://gwb.readthedocs.io). 

Each page of the online manual contain a small todo box of what content the page should contain. This is usually just one or two short sentences. In some cases the page contains a bit more than that, but that is mostly to show what kind of functionality can be expected.

Any feedback is welcome, but I have some question in particular:

1. Is any important information missing from the manual?
2. Should any pages be split up or merged?
3. Should information/pages be ordered in a different way?
4. Is there information or are there pages which should not be in this manual?
5. Do you have any concerns with this format in general?

Feedback can also be provided by directly contacting me. Also please see the tracking issue for the new manual #379. Futhermore, feel free to directly make pull request on the manual. Pull request on pages can be directly made from the manual interface (see the github icon in the top of each manual page).

I will wait at least a week with adding significant content to the manual to provide an opportunity for feedback and discussion if needed.

MFraters:
@naliboff and @bangerth, thanks for going through the manual outline so quickly! Both your comments have been helpful and are really appreciated!

@naliboff, I agree that there seems to be a bit of redundancy in the developer manual section. I really tried to separate out abstract concepts, mixed abstract practical items and practical gwb specific items. I may have taking this too far, but I agree that that would be easy to change later by remixing (as opposed to rewriting) the text. So I will try to do it this way for now and will change it if it becomes this style is more confusing than helpful.

@bangerth, I have made an attempt to address your comments in pull request #387.

PR #387, opened 2022-02-10
Address new manual outline first round of comments.

MFraters:
This pull request addresses some of the issues with the new manual outline raised by @bangerth. I read your comments as  the two following issues (please correct me if I am wrong)

1. One issue is that there is not a good introduction of what the world builder is and the philosophy behind it.
2. I didn't explain have a section on the API and did not realize in the manual that that is a completely different audience.

For the second item, I think that there are now actually three different audiences: people who wat to make input files, people how want to link the GWB to their code and people who want to modify and/or extend the GWB. I have now put them into three different sections and made a global introduction section where I put the items of point 1 into.

@bangerth Do these changes to the outline address your issues?

The preview of the manual with these changes can be found here: https://gwb--387.org.readthedocs.build/en/387/

Related to #379  and #386.

### DIFF
--- a/doc/sphinx/api_manual/API_FFI/CPP_interface/ASPECT_integration.md
+++ b/doc/sphinx/api_manual/API_FFI/CPP_interface/ASPECT_integration.md
@@ -1,0 +1,6 @@
+ASPECT intgration
+=================
+
+```{todo}
+Explain how the GWB is integrated into ASPECT
+```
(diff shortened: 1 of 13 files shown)


MFraters:
Thanks for the reply Wolfgang! Sound like a good plan to me.

Thanks for your the comments @tjhei! I think this pull request addresses the issue in the way you suggested. You can find a live version of the manual in the pull request here: https://gwb--387.org.readthedocs.build/en/387/

tjhei:
Looks much better, yes. :+1:

PR #392, opened 2022-02-20
Switch unit tests to doctest

MFraters:
I ran into some issues with catch2 with a new version of gcc. Updating catch2 did solve the issue, but it made me look at what the current options are. One of the things which started to become more of an issue is the compile times of the unit tests. Doctest promises to reduce those compile times significantly. On my laptop a single line change in the main unit test file took about 14s with catch2, doctest reduced it to 7s and with [super fast asserts](https://github.com/doctest/doctest/blob/master/doc/markdown/configuration.md#doctest_config_super_fast_asserts) it is down to 4. I am mostly hoping it will speed up the testers, but it is also a difference I will notice in production.

For this setup is can nearly be used as a drop-in replacement. The only real issues I encountered was that there was no option to use a Contains check for the CHECK_THROWS_WITH. That is now hopefully being solved with https://github.com/doctest/doctest/pull/620. Since that works great for this project, I have used the header file produced by that pull request.

This pull request could have been made with even less changes, but I took the opportunity to also combine all the unit tests into one binary. If desired, the unit tests could in the future actually be integrated directly into the code or headers, but I am not sure yet that would be an improvement for this library.

I also had to add the limits header to gml.h. I don't know why that was not needed before.


MFraters:
The whole testsuite went from 9 and a half minutes (from a pull request a week ago) to 5 and a half minutes without loss of functionality, and the potential for new functionality. That seems like a significant improvement to me. 

The pull request with the changes for the doctest header have not been merged yet in the doctest repository and the review process may still change that code, but since this speeds up the development I think it should be merged.

Issue #366, opened 2021-11-06
Make min and max depth variable on postion in plate.


MFraters:
Currently only fixed values are allowed for the min and max depth in the world builder. Instead of that single value (e.g. 100e3), I propose to change this to define instead a array of objects: [{100e3}] to produce the same results. This would internally expand to [{100e3,[[0,0]]}]. The first number is the depth and the second double array is a list of points with that depth. A more complete example would be: [{100e3,[[10,20],[20,40],[80,60]]},{20e3,[[20,60],[10,30]]}]

The idea of this would be that the depth would be interpolated over multiple points. If there is only one point, that will be use everywhere. If there are multiple, the value (the first number) will be interpolated based on the distance to all the defined points. I currently think this is going to be the cleanest way to implement this. This will be a significant breaking change for the input file and will most likely break all input files. This implementation will however be minimal overhead if you are not interested in the feature, but allow for a huge amount of flexibility in thickness (top and bottom) of plates and for examples layers in plates.

Making a passive margin or a thickened crust would be very easy with this approach. Because there is no connectivity, it may become expensive if a large amount of points are involved. If this would become an issue, a mesh could be generated during initialization, which would make looking up the values a lot faster. 

I will have to do a lot more thinking on the implementation of the interpolation, but I do like the new proposed syntax. If there are better or different proposed ways for the syntax, or ideas how to do the interpolation, feel free to post here.

P.S. I have already made an initial post about it in #132, but I now realize that it would have been better to make this a separate issue.

MFraters:
I have been looking a bit more into this and `[{100e3,[[0,0]]}]` (array containing object) need to become `[[100e3,[[0,0]]]]` (double array, containing a double in the first entry and a double array in the second entry). The difficulty comes from JSON requiring a key for all the values in an object. JSON doesn't require all entries of an array to be the same type.

Technically we could probably allow both `"min depth": 100e3"`,  `"min depth": [[100e3]]"` and  `"min depth": [[100e3,[[0,0]]]]"` at the same time (with the `OneOf` statement in the schema). This would allow for backward compatibility, but I am worries having multiple allowed syntaxes which are clearly different (`"min depth": 100e3"` vs  `"min depth": [[100e3]]")`, and not just a secret default value (for example `"min depth": [[100e3]]"` vs `"min depth": [[100e3,[[0,0]]]]"`) will lead to more confusion with users.

[...]

PR #394, opened 2022-02-23
Add kd tree as first step to variable plate depth

MFraters:
This is the first part of add the option for min and max depth variable on position in plate (#366). This pull request add a compact kd-tree. It is based on the algorithm as implemented in the Templatized Geometry Library (TGL) (https://web.cs.ucdavis.edu/~okreylos/ResDev/Geometry/index.html) as send to me by the creator Oliver Kreylos @Doc-Ok, but re-implemented from scratch optimized for a very specific purpose. 

The unit tests are based on hand drawn examples I made before implementing the examples. I had a slightly different indexing scheme (if there is an even number of points, choose the one with the highest x/y), which for an odd number of points results in the same result node vector, but for even numbers there is a difference in the order. The algorithm of Oliver has the advantage that indexing is simpler, cheaper (single bitwise operation) and seems to be more contiguous to me when transversing the tree, so I decided to go with that.

The most important thing which is missing is a function which return the closest points. TGL contains such a function, but it is not clear to me what it exactly returns, so I have to look a bit more into that. 

@Doc-Ok, can you confirm that this is ok to use these algorithms in the world builder (LGPL license)? I will be happy to add you as contributor in the AUTHORS file.

### DIFF
--- a/include/world_builder/kd_tree.h
+++ b/include/world_builder/kd_tree.h
@@ -1,0 +1,156 @@
+/*
+  Copyright (C) 2022 by the authors of the World Builder code.
+
+  This file is part of the World Builder.
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Lesser General Public License as published
+   by the Free Software Foundation, either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


MFraters:
Received confirmation that the licensing is oke.

PR #396, opened 2022-03-23
Add variable surface min max area features

### LINKED ISSUE #366: Make min and max depth variable on postion in plate.

MFraters:
This pull request main objective is to resolve #366. It is now also added in a backward compatible way. That means that for example `"min depth":0` and `min depth":[[0]]` are both valid and will give the same result. 

The pull request currently only adds this functionality for continental plates, but adding it for oceanic plates and mantle layers should be easy.

The work on this pull request raised some issues along the way which had to be, or where a lot more convenient to resolve directly. This means this is a big pull request, which I am doubting about splitting up. For now I am keeping this together and I will reorganize the commits.

Here are some pictures of what is now possible with this input file:
```
{
"version":"0.5",
"coordinate system":{"model":"cartesian"},
"features":[
  {
    "model":"continental plate", "name":"Var thickness continent", "coordinates":[[0,1000e3],[1000e3,1000e3],[1000e3,0],[0,0]],
    "min depth":[[50e3,[[0,500e3]]],[200e3,[[1000e3,500e3]]]], "max depth":[[300e3],[200e3,[[0,500e3],[1000e3,500e3]]]],
    "temperature models":[{"model":"uniform", "min depth":0, "max depth":500e3, "temperature":150}],
    "composition models":[
      {"model":"uniform", "min depth":[[50e3,[[0,500e3]]],[200e3,[[1000e3,500e3]]]], "max depth":[[200e3],[100e3,[[0,500e3],[1000e3,500e3],[0,0]]]], "compositions":[0]},
      {"model":"uniform", "min depth":[[200e3],[100e3,[[0,500e3],[1000e3,500e3],[0,0]]]], "max depth":[[300e3],[200e3,[[0,500e3],[1000e3,500e3]]]], "compositions":[1]}
    ]
  }
]
}
```
variiable_thickness_continent_v2_back
variiable_thickness_continent_v2_front


MFraters:
Spherical now also seems to work, here is an example figure of what that looks like:

variable_thickness_continent_spherical

(basically the same input file, but with spherical coordinates:
```
{
"version":"0.5",
"coordinate system":{"model":"spherical", "depth method":"begin at end segment"},
"features":[
  {
    "model":"continental plate", "name":"Var thickness continent", "coordinates":[[0,10],[10,10],[10,0],[0,0]],
    "min depth":[[50e3,[[0,5]]],[200e3,[[10,5]]]], "max depth":[[300e3],[200e3,[[0,5],[10,5]]]],
    "temperature models":[
      {"model":"uniform", "min depth":[[50e3,[[0,5]]],[200e3,[[10,5]]]], "max depth":[[200e3],[100e3,[[0,5],[10,5],[0,0]]]], "temperature":150},
      {"model":"uniform", "min depth":[[200e3],[100e3,[[0,5],[10,5],[0,0]]]], "max depth":[[300e3],[200e3,[[0,5],[10,5]]]], "temperature":200}
    ],
    "composition models":[
      {"model":"uniform", "min depth":[[50e3,[[0,5]]],[200e3,[[10,5]]]], "max depth":[[200e3],[100e3,[[0,5],[10,5],[0,0]]]], "compositions":[0]},
      {"model":"uniform", "min depth":[[200e3],[100e3,[[0,5],[10,5],[0,0]]]], "max depth":[[300e3],[200e3,[[0,5],[10,5]]]], "compositions":[1]}
    ],
  }
]
}

MFraters:
In terms of functionality and basic testing, this pull request is mostly done. Benchmarks also show that  also shows that there is no significant performance loss if this feature is not used, which is great. 

I have a few more corner case tests and a benchmark in mind and will need to improve the coverage. Will also need to properly document all the new and changed functions. After that the code and commits will be cleaned and refactored. Documentation into the user manual will be part of #379.

MFraters:
This was a large amount of work, but I am happy with how this pull request turned out. The thing left is to add the same feature to the oceanic plate and mantle layer and add similar tests. I expect this pull request to be merged within the next few days if there are no further issues or comments.

Additional benchmarks can only be added in a follow up pull request because main doesn't have the features yet.

PR #401, opened 2022-03-29
Remove non contiuous interplation

MFraters:
This is replacing #316 in order for the pull request to be more systematic. The goal is to get rid of the different interpolation schemes for the slabs (and in general). In short: the old scheme has several issues and the contentious method is generally  better, so I don't see a reason it should remain available. The only advantage is that the old method is slightly faster, but that can potentially be mostly mitigated with the new optimizations which will become possible by removing the old method.

- [X] Change the default to continuous monotone spline and update tests. This way the pull request has a good basis to test other changes against.
- [X] Set interpolation code in the tester to only increase with steps of 1 and update the tester. (done in #400)
- [X] Remove code which is not used anymore.
- [X] Optimize the interpolation code
- [X] Optimize the rest of the code.
- [X] Add changlog entry


MFraters:
This pull request was a lot more work then expected, especially to get a good fast and stable convergence for the closest point on spline algorithm. I also found and fixed a lot of corner cases and improved the overall accuracy (see images below).

difficult_curved_slab
difficult_curved_slab_spherical_right_new


Performance wise I this this pull request is in a good place, and I only expect some minor tweaks. At this point I will focus on checking and updating the tests, cleaning and documenting the code and make a change log entry. When this pull request is done, the new interface can be made and several more global optimizations can be made so that this code is called less often.

MFraters:
This pull request is ready to merge. It is well covered, the run time is of the "correct" method is ~-40% and compared to the "old" method the benchmark show 0-16% more time taken, which is acceptable for now (down from almost 80% more time).

This pull request has also fixed a lot of corner cases for the new/correct method. I also decided to add and change many unit tests to approval test. Approval test are a hybrid between classic unit tests and integration test. It allows you to test individual functions, but store the results automatically in a file to compare against a "golden standard". I decided to do this because of the improved accuracy of the new method I would have to update more than 700 unit tests check (probably several times) line-by-line by hand, which would have been extremely time consuming. This method allows for easier updates of the results if needed.

I have done some cleaning of the git history, but the commits of this pull request could be cleaned up more than I currently have done. I decided not to clean it up more than it currently is, because it would be a lot of work and I do value the remaining history here, because it shows the different algorithms I have tried to solve the closest point to spline problem.

PR #408, opened 2022-04-15
new surface bounding box optimalizations.

MFraters:
Now that the radius of the sphere is defined, the surface_bounding_box can be fully precomputed.


MFraters:
@alarshi @bangerth This may be of interest to the both of you. In the benchmarks here it makes a small but (just) significant difference. It also show the newly opened up the possibility of more optimizations as for example proposed in #308, now that the spherical radius is known.

PR #421, opened 2022-04-27
[WIP TEST] Astyle 3.1

MFraters:
This pull request is a test to see the differences between astye 2.04 and astyle 3.1. I am currently not thinking of switching before the next release (and maybe not even after), but I was curious to see how different it would be and provide an easy way to view and discuss the differences. 

The pull request is divided into two commits. The first commit contains the persistent changes (2.04->3.1->2.04). The second commit contains the non-persistent changes.

There are not many changes (60 lines in total). A large part of the changes in both commits are in the doctest library and the rapidjson library, which maybe should not even be part of the indent (but that is a separate discussion I guess). If you exclude those libraries there are 32 lines which change within the world builder, of which 16 are persistent.


MFraters:
Forgot to say that his came up in #420

bangerth:
Yes, this really isn't very much. It would be interesting to see how that looks like for ASPECT.

I also think you're right that you should exclude imported projects from indentation. These ought to be preserved in exactly the state they were imported in.

PR #423, opened 2022-05-12
added age dependence to depth gradient of minimum slab temperature

mibillen:
For very young slabs, need to allow the minimum slab temperature to increase more rapidly with depth. I had originally planned to have age dependence but I didn't have enough constraints from the citcom models to know what this should be. Here, I created an age factor that is added to the subduction-velocity factor. The age factor starts to have an effect for slab younger than 40 my and it linearly increases the depth gradient of the minimum slab temperature so younger slabs warm up faster than older slabs.

### DIFF
--- a/source/features/subducting_plate_models/temperature/mass_conserving.cc
+++ b/source/features/subducting_plate_models/temperature/mass_conserving.cc
@@ -330,11 +330,17 @@ namespace WorldBuilder
 
               double zero = 0.0;
               double one = 1.0;
+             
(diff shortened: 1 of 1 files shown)


MFraters:
Thanks for the quick fix! The solution sounds good to me. Can you indent the code and add a test for a young slab?

MFraters:
Thanks for the quick fix! I am testing it now. Can you indent the code and add a test?

MFraters:
closed by #470

PR #425, opened 2022-05-16
Remove unused variables (and avoid corresponding warnings.

bangerth:


MFraters (approved):
These where leftovers forom #401. Thanks for cleaning it up!

PR #426, opened 2022-05-16
Avoid warning about missing braces {...}.

bangerth:


bangerth:
Compilers seem to interpret the standard in different ways, and my compiler gives me a warning about another set of missing braces `{...}` which I can quell by adding them. Let's see whether your tester accepts this. 

We have had these sorts of difficulties in deal.II as well, and usually adding another set of braces works.

MFraters (approved):
I don't see any warnings for the linux and osx testers, so I am happy to merge this.

PR #430, opened 2022-05-23
Add change log entry for smooth fault composition.

alarshi:
This PR adds a change log entry for PR#356.


MFraters (approved):
Thanks, looks good!

PR #431, opened 2022-06-16
Add content to manual introduction part

MFraters:
adding content to page `What is the GWB`.

Related to #379

### DIFF
--- a/doc/sphinx/index.md
+++ b/doc/sphinx/index.md
@@ -19,7 +19,7 @@ introduction/bibliography
 
 
 ```{toctree}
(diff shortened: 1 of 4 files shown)


MFraters:
I added content for all the pages of the introduction (3 pages without the bibliography), since I think it makes more sense to make pull requests for a few pages at a time, instead of for every single page. 

@bangerth and @naliboff, I remember you where interested in reviewing the new manual sections. I will wait a few days with merging, if you need more time or don't have time to read at all, let me know :)

PR #436, opened 2022-06-22
rtd sphinx fail on warnings.

MFraters:
I would like read the docs to fail on warnings, but still build. There should currently be a warning, so read the docs should return that is failed, but still build the documentation: https://github.com/readthedocs/readthedocs.org/pull/7251


MFraters:
related to #379

PR #443, opened 2022-06-28
[WIP] Rename and move source files and folders

### LINKED ISSUE #442: Move lib and apps source file directories
As a follow up to #378, the source folders of the app and visualizer need to be renamed. I have not been happy with the sources for the apps living outside the source folder, so this would be a good time to change that. 

I am proposing putting what is currently in `source/` into `source/world_builder/`, and then add a folder in the source directory for both `gwb-dat` and `gwb-grid`. For the include files, I would  only rename the `app` and `visualizer` folders to their new names. I doubted about naming the folder for the current files in source to gwb-lib, but that would then also require the currently include/world_builder folder to be renamed (if we want to remain consistent), which would definitely break other codes like aspect and the name in the include would be less clear.

MFraters:
Major change in location of all the source files. See #442 for discussion.

Closes #442.


CMakeLists.txt
```diff
@@ -259,8 +259,8 @@ ELSE()
   MESSAGE(STATUS "Disabling unity build.")
 ENDIF()
 
-add_executable(gwb-dat "${CMAKE_CURRENT_SOURCE_DIR}/app/main.cc")
-add_executable(gwb-grid "${CMAKE_CURRENT_SOURCE_DIR}/visualization/main.cc")
+add_executable(gwb-dat "${CMAKE_CURRENT_SOURCE_DIR}/source/gwb-dat/main.cc")
+add_executable(gwb-grid "${CMAKE_CURRENT_SOURCE_DIR}/source/gwb-grid/main.cc")
```
bangerth, line 263: I think it makes sense to have source files in directories that match the executable's name.

PR #445, opened 2022-07-20
Add WORLD_BUILDER_VERSION_GTE macro to config.h

MFraters:
This add a macro function `WORLD_BUILDER_VERSION_GTE(major,minor,patch)`, which works the same as the  deal.ii macro function `DEAL_II_VERSION_GTE(major,minor,subminor)` (see https://github.com/dealii/dealii/blob/38cf3f584366bb27d2d94dd994675011842e8fac/include/deal.II/base/config.h.in). I think 100 patch version and 100 minor releases per major version should be sufficient for the time being... 

It doesn't do anything with the label (`-pre,-rc1,etc.`), but that is probably for the better since I think pre-releases should be treated as a full release of that version.

related to #444. @bangerth

### DIFF
--- a/include/world_builder/config.h.in
+++ b/include/world_builder/config.h.in
@@ -28,6 +28,19 @@
 #define WORLD_BUILDER_VERSION_PATCH @WORLD_BUILDER_VERSION_PATCH@
 #define WORLD_BUILDER_VERSION_LABEL @WORLD_BUILDER_VERSION_LABEL@
 
+/**
+ * Returns true if the used World Builder version is greater or equal than the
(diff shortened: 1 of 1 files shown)


MFraters:
Thanks. I would suggest to leave it as it is for now in aspect. Users could theoretically link their own gwb 0.4.0 instead of the version which is shipped with aspect (although I see no reason why they would do so). Maybe at some point we can require at least 0.5.0.

bangerth:
Right. What I meant is that we create the same kind of macro (just prefixed with `ASPECT_`) in ASPECT, and then we can use it there as well. But that's neither urgent nor important.

PR #447, opened 2022-10-04
Fix mass conserving temperature.

MFraters:
@mibillen I got this patch from you a while ago, which if I remember correctly fixes the mass conservative temperature for very young slabs. Could you take a look at this if this is correct?


MFraters:
closed by #470

Issue #446, opened 2022-07-28
Input dip point in degrees using spherical coordinate system


alarshi:
When using dipping faults in WB for the global mantle flow models, I realized that the `dip point` in the JSON file (internally used as `reference_point` in WB) needs to be in radians and not degrees as the longitudes/latitudes of the fault trace. 
Maybe for consistency, we would want to give the `dip point` in degrees when using the spherical coordinate system.

MFraters:
Thanks for opening an issue about this. I agree that it makes more sense to have the dip point in degrees. I will make a fix for that.

PR #448, opened 2022-10-06
Modify the handling of dip point to degrees.

alarshi:
This PR addresses the issue #446, so that we can input "dip point" parameter in degrees as with the coordinates.


MFraters:
Great, thanks! I realized that since it might be a breaking change/fix that it would be useful to also add a change log entry in the Fixed category. Could you add that as well?

MFraters:
Thanks!

Issue #449, opened 2022-11-04
Add Bezier curve


MFraters:
Based on the work done for #429, I found that using Bezier curves should be a lot more reliable and faster since it has a closed form solution. I think the Bezier curves should be added for slab and fault coordintes and maybe even replace the continuous monotone spline.

MFraters:
Here is a comparison between the Contiuous Monotone Spline (CMS) and the Bezier curve for a 3d spherical subducting slab setting.

![3d_spherical_slab_depth_temp_contour_CMS-vs-Bezier_above](https://user-images.githubusercontent.com/7631629/200497360-7e84a5e7-e660-4afe-9822-cc5bcd2ab8e3.png)
![3d_spherical_slab_depth_temp_contour_CMS-vs-Bezier_below](https://user-images.githubusercontent.com/7631629/200497363-8f9f6d7e-3c59-4e59-8bb3-5f6c0512564e.png)


 You can argue over which is better, but the Bezier curve approach is definitely smoother. I personally think the Bezier approach is both numerically (more stable, no iterations which can fail and should be faster) and visually better. The CMS follows the points with less curvature, but if you really want that you can just add more points.

Looking at these results I now more strongly think that the CMS approach should be replaced by the Bezier curve approach, unless major drawbacks are found in the mean time of course. Although both could be supported, I don't see much of an advantage and it would make the code a lot more complicated and harder to maintain (and also allow users to get hard errors that an iteration failed).

May be of interest to @alarshi, @danieldouglas92 since it will (I think only slightly) change initial conditions compared to the current main branch.

danieldouglas92:
I also agree that the Bezier curve does look better! I'm excited to try it once it's implemented

Issue #429, opened 2022-05-20
[Tracking issue] slab2 integration


MFraters:
This issue is to track the development of the slab2 integration. 

The original idea was to use the 2D spline information from slab2 and use that. The advantage is that is the most accurate accurate representation of the slab2 data possible. The downside is that it is practically impossible to tune and tweak the geometry, because it is made out of lots (thousands?) of spline coefficients. 

The new idea is to use the contours of the slab2 data, which is provided on their website. The downside of this is that it is not a perfect representation of the slab2 splines. In reality it will be very close though, since we will also use splines to interpolate the data. The advantage is that this can be represented in a human readable and writable way. 

The input format could look something like this: `[[50e3,[[0,0],[10,10],[10,20]]],[100e3,[[0,10],[20,20],[20,30]]]]`. This input doesn't account for how to input different temperature and compositional models per coordinate, and for example, different thicknesses, so what are now the coordinates may become an object.

To make a fully smooth surface for the slab/fault we need either a spline or the dip at every point. Computing the 2D spline would probably have to be done at the fly and would be very expensive. If we would have a dip at every point, we interpolate that with the closest point on the spline position and then get a dip for the top and bottom and do a cubic interpolation, which would create a smooth surface in 3D.

may be of interest for @naliboff, @mibillen, @jdannberg, @alarshi, @danieldouglas92, @bangerth

PR #452, opened 2022-11-08
[WIP] Add bezier curve

MFraters:
This is part of #429 and #449. This pull request removes the Continuous monotone spline and replaces it with Bezier curves. See #449 for the rationale. 

The code is very much in a wip state, but it is function. Basically all fault and subucting slab test results need to be updated.


MFraters:
I wanted to give an update to address and document some of the issues and considerations I have been facing.

First of all, the analytic solution works only in Cartesian coordinates and it is slower than I hoped for (but still mostly a significant improvement). So I set out try again a Newton solver approach with a potential fall back to the analytic approach. This worked reasonably well, but I found some issues. The main issue is actually one which is is also an issue in the normal case, but expresses itself more with the Newton solver approach. The problem can be reproduced with the following parameter file (focus on the coordinates, the rest is not that relevant for this issue): 
```
{
  "version":"0.5",
  "cross section":[[0,50e3],[50e3,0]],
  "interpolation":"continuous monotone spline",
  "features":
  [
     {
       "model":"subducting plate", "name":"Antilles slab", "dip point":[1e7,1e7],
       "coordinates":[[50e3,12e3],[125e3,75e3],[80e3,25e3],[200e3,-10e3]], 
       "segments":
       [
         {"length":350e3, "thickness":[100e3], "angle":[0,45]}
       ],
       "temperature models":[{"model":"uniform", "temperature":600}],
       "composition models":[{"model":"uniform", "compositions":[0,1], "fractions":[0.25,0.75]}]
    }
  ]
}
``` 
And here is a figure of what the trench looks like in this case (the reference point is up in the figure): 
problem_trench_geometry

The problem occurs around the sharp point. As an example, below is the line between a check point and a computed closest point on line near this sharp point:
problem_trench_geometry_check_point

On this scale, there doesn't appear to be a real issue, since you should just be able to create a slab from this. Unfortunately, this is not the case. To get to the core of the issue, how the actual Bezier curve is generated. It starts with making a cubic spline through the points. Because I would like to use quadratic Bezier curves (they should be simpler and should be faster), I interpolate the center point of the spline and make these the control points (I compute and store the angles which make these points, but that is for different reasons). The point where this breaks down is actually with the splines themself. To visualize this, take a look at this site: https://mbostock.github.io/protovis/ex/splines.html

I have made some screenshots:
spline_test_1
spline_test_3
spline_test_2

When I use the actual numbers for the Bezier curves I generate through this site: https://www.desmos.com/calculator/lvdgnyhkvy, I get the following: 
bezier_1and2 1

This probably explains why the Newton solver is having such a hard time.  It is a extreme version of the second spline screenshots, where both lines say the closest point is on the other line. 

I can see two ways of fixing the issue, besides ignoring it, because it is kind of an unrealistic corner case:

1. Don't use a spline as a basis. That would mean I would need to use a different method to automatically compute reasonable control points. 
2. Detect the issue and move the control points. 

I think the second approach is feasible, and I am going to try that. The basic idea is that you can check on which side of a line a control point is. If you do that for the line between the first and second and the line between the second and third point, then you can check if they are on a different side of the line (which I think they should be). If they are on the same side, then the point should be mirrored over to the other side of the line. I might be wrong about this, but I thought I needed to document this, so it is easier to take further steps forward.

MFraters:
closed by #462

PR #453, opened 2022-12-08
interface for the distance to a feature's plane

lhy11009:
I added a simple interface that could be used by ASPECT to call the distance_point_from_curved_planes function for computing the distance to a feature's plane surface.
I have also added a test in the unit tests for testing the implementation of the interface.

PR #454, opened 2022-12-19
add openmpi-dev as an install to the testers.

MFraters:
It looks like the mpi tester is failing due to it not finding it as noted in #453. This pull request attempts to fix that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -397,7 +397,7 @@ jobs:
       run:   |
          if [ "$RUNNER_OS" == "Linux" ]; then
             sudo apt update;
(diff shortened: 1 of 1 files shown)


MFraters:
The problem has to do with the ubuntu-latest being changed to 22.04: https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/ and something with openmpi changed. Before just installing openmpi-bin was needed, but now also libopenmpi-dev is required. Furthermore, cmake 3.1.x doesn't seem to work on ubuntu 22.04.

PR #462, opened 2023-01-24
Add bezier curves

### LINKED ISSUE #449: Add Bezier curve

MFraters:
Builds on and replaces of pull request #452, resolves #449 and is a step towards #429. This replaces the Continuous Monotone spline for cubic Bezier curves. 

In #452 I wanted to use a quadratic curve since it can be solved analytically, but it turns out that the derivatives between the segments of the spline are not necessarily continuous, which leads to issues. With the cubic splines, you can construct them to have continuous derivatives (although not second derivatives, but that is fine). 

The implementation can either automatically construct the angles of the start and end of the segments, or they can be provided to the function. For now I am not exposing the option for the user to provide these custom angles. I am not sure if that has enough value for trenches and faults, but it will certainly be used for the angles of the slabs. 

I have put quite a bit of work in the line search algorithm, and I think it actually if a lot better for this application than a normal line search algorithm based on the testing I have done. 

This pull request still requires some clean up and updating of the tests, but I think this approach is better than the continuous monotone spline, especially because it allows unifying algorithms an code with slabs(2) and faults.


MFraters:
One thing I forgot to mention is that this pull request also changes how the reference point interacts with the direction of the slab/fault. It used to be calculated based on on which side the reference point is compared the line which the closest point is on. This unfortunately creates issues with trenches which bend a lot, causing the slab/fault direction to change. Since I don't see how this is useful for anyone, and the same can be achieved intentionally, it will now be based on the line between the first and last point of the trench. This is a lot more intuitive for the user.

MFraters:
This pull request is now ready to merge. Since it changes quite a fundamental part of the world builder, I will wait with merging until at least the end of next Monday, to do some more testing myself and give others the opportunity to test,. Thanks to @danieldouglas92 for all the help with testing this and finding issues during the development. In the end it should not change much to what the model looks like as much as I originally thought it would, but you may still see some significant differences. 

@danieldouglas92, @alarshi, @rfildes, if you want to feel free to give it a go and let me know if you run into any issues.  Or if you want to test it, but need more time, let me know.

PR #466, opened 2023-01-27
Add gdb and valgrind testing

MFraters:
Adding gdb and valgrind as extra testers will help finding problems faster and more easily. This allow to run the tester with either or both gdb and valgrind. This pull requests also adds a github action tester which tests both of them. They fail if they find an issue.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -436,3 +436,75 @@ jobs:
       working-directory: ${{github.workspace}}/build
       shell: bash
       run: ctest -VV -C ${{ matrix.build_type }}
+
+
+  linux-gdb-valgrind:
+    needs: [test_indentation]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-22.04']
+        build_type: ['Debug']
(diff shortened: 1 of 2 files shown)


MFraters:
@tjhei and @gassmoeller Could something like this (maybe more limited?) be useful for aspect as well? The pre-building is taking most of the time compared to the running for aspect, so maybe the overhead is not that large (although valgrind can be really slow)

tjhei:
The only reason to run under gdb is to get a nicer call stack when a crash happens, right?

MFraters:
Yes, it will give a bit more info when failing (usually also some variable values). I also have the feeling thta i sometimes catches things more early, but I am not sure if hat is true. For the gwb, to run both the normal tester and the gdb tester in addition takes less then a minute (compared to less than 20 seconds normally), so I am considering turning it on for all platforms that have gdb. Valgrind is definitely more expensive, it takes about 9 minutes with the current settings. If you use the `--track-origins=yes` as well, which gives some more details on the problems it finds, it takes about 15 minutes.

PR #457, opened 2023-01-12
Young slabs

MFraters:
Follow up on #423. @mibillen asked me to follow up on this and I forgot about it. I can't push to her branch so I am making a new pull request. All the work is by @mibillen!

PR #470, opened 2023-02-01
Smooth depth dependent slab temperatures

mibillen:
I have completely reformulating the empirical models for the minimum slab temperature and its offset from the slab surface. 1. Instead of using linearly increases with depth, I'm using a complimentary error function. This allows the slope of changes to be larger at the coupling depth and much smaller either near the surface or much deeper in the slab. I also changed the taper for the slab tip to also use an erf function. 
2. The empirical model still uses factors to adjust the width and height of the erfc functions depending on the subducting plate velocity and age of the plate.  I have used min/max values to prevent these from being non-sensical numbers. 
3.  Tmin and offset now depend on depth of the slab surface rather than distance along the slab, and the values have been corrected to properly include the same adiabatic gradient as in world-builder. This change means that the parameter "shallow slab dip" is no longer necessary and has been removed. 
4. Note that the effective age of the plate still increases with distance (as it should). The slab tip taper also still depends on the distance within the slab tip.   
5. Finally, I have added a parameter to control cooling of the forearc that occurs over time (but is not captured by the mass conserving model). I have added an input parameter to allow the user to adjust the amount of cooling... but note that increasing this value VIOLATES the mass conservation assumption and does affect the deeper slab (play with the value and you will see). However, this value is needed to insure that even very young plates develop a thin thermal boundary layer on top.  This new parameters "forearc cooling factor" has a default value of 1.0. So, you only need to add this factor in if you need additional cooling of the fore-arc. Note that  this cooling should be minimal for young plates, but could be substantial in subduction zones that have been active for a long time with old slabs, as the thick, cold subducting plate insulates the forearc.


MFraters (commented):
My initial tests with it look very good, but I should have more data tomorrow. I like that the code is well commented, just one small request for a clarification. Other than that, please update the test results when we agree that this is the final version. 

This replaces and closes #457 and #447 and #423

@danieldouglas92 and @rfildes, may be good to test out.

MFraters (approved):
Thanks for all the effort! Looks good. Can you add a change log (in CHANGELOG.md) entry about the changes to the mass conserving temperature model?

PR #480, opened 2023-04-13
Only read world build file on one MPI rank

gassmoeller:
Same PR as geodynamics/aspect#5132. This avoids opening the world builder input file on every MPI rank. Instead it is only opened and read on the first rank and then broadcast via MPI to all processes. This is very useful for large-scale run on HPC clusters, because the file system can easily get overwhelmed when tens of thousands of MPI processes access the same file.


MFraters (approved):
Thanks for making this pull request, it is very much appreciated! 

The code looks clean, reusable and well documented and it works for all the tests both here and in aspect. Can you add a changelog entry? Don't worry about the failing documentation tester or the test that is not picked up by github actions. They are adressed in #481 and #482 respectively.

gassmoeller:
Changelog entry added.

MFraters:
The loss in coverage is due to error catching code, which is fine.

PR #481, opened 2023-04-13
Remove extra_navbar from sphinx conf.py

MFraters:
`extra_navbar` was removed in https://github.com/executablebooks/sphinx-book-theme/pull/640, and now throws a warning. Since warning are turned in to errors by the settings, we probably need to remove this setting. I will see how the results look like online and see if it need tweaks.


MFraters:
The fix works and the updated theme (which caused the issue) looks good to me.

PR #484, opened 2023-05-09
Rounded fault

alarshi:
I am pushing this branch which we we have been using for our global mantle flow models. This PR is only to get the relevant commits from this branch merged into current main.

@gassmoeller : The commits using the rounded corners and interpolation between two very closely spaced faults are : 3e888173d4a73
d18d1c404a01
e33e739ea7560


MFraters:
Hey @alarshi,

Thanks that you are willing to contribute this work back to the world builder! I can take a closer look at the code when I get back from traveling. Could you rebase in the meantime and squash your commits down a bit? Preferably in logical blocks. Thanks!

tjhei:
It would be nice if you could rebase to the latest GWB main version as it is quite difficult to configure this current version with the current ASPECT master.

alarshi:
@tjhei : Yes, I agree that this version is not compatible with the main ASPECT and I have been working around this with the old CMakeLists. However, I was not able to rebase this version on the main WB either because a lot of the core code has changed/restructured on which the commits of this branch were based of.  I discussed with @MFraters on leaving this branch here for reference and getting the functionality in WB as separate PRs. I still need to figure out the differences between the main and this branch results for our models. This is one of my goals for this hackathon.

PR #485, opened 2023-05-11
Avoid accidental vector copies in inner loop

gassmoeller:
Whenever the `properties` function of a feature is called it currently creates a `const` copy of the properties vector. There is no reason to do this (if it is `const` it doesnt matter if we read the copy or the original). This function is called in a loop over all features and in our application we have about 10,000 features, so for every evaluation point we currently allocate and deallocate 10000 vectors. This little change saved about 25% of the total WorldBuilder time in a small example model.


MFraters (approved):
This is great, thanks @gassmoeller!

This is clearly a significant improvement, which can be noticed in the benchmarks (although no 25% in those benchmarks). The tester is also happy and you added a nice change log entry, so I see no reason not it merge it :)

I am wondering if it makes sense to add your test to the benchmarks which are automatically run. What do you think?

PR #486, opened 2023-05-11
Inline a heavily used function.

gassmoeller:
This function is used frequently in `Features::Fault::properties` and is pretty short. Inlining it significantly reduces the cost of the function (from about 30 to ~5 instructions) and saves on the order of 1% of the world builder time in my model with many faults. Not much, but the change is also small.

### DIFF
--- a/include/world_builder/objects/natural_coordinate.h
+++ b/include/world_builder/objects/natural_coordinate.h
@@ -93,6 +93,32 @@ namespace WorldBuilder
          */
         std::array<double,3> coordinates;
     };
+
+
+    inline
+    double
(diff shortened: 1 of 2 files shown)


MFraters (commented):
I am a bit more hesitant with this one. The benchmarks do not really show a significant difference, if anything a small slowdown in some of the benchmarks. I guess it may be compiler dependent? 

I might still be fine with the change, but lets talk about it on Monday.

tjhei:
Are the benchmarks run in release or debug mode?

gassmoeller:
Benchmarks are run in release mode. Not sure what to make of this PR anymore.

PR #487, opened 2023-05-11
[WIP] Optimize bezier curve find closest point

gassmoeller:
This is probably the most controversial of my changes and needs further testing.

All in all this speeds up the `closest_point_on_curve_segment` function by about a factor of 15 by two changes:
- A factor of 5 is contributed by the switch to approximate fast sine and cosine functions. This should be less problematic as long as the Newton iteration converges properly, because we are dealing with guesses for points most of the time anyway (do I need to use the accurate sin/cos function to compute the final residual in the newton iteration?)
- A factor of 3 can be gained by a change to the algorithm. Instead of running a Newton iteration on every bezier segment I can use the estimated point to find the the most likely closest segment first, then determine the closest  point with the newton iteration only on the closest segment. There is theoretically a risk that the estimated point is not accurate enough to find the closest segment. This will probably depend on how "smooth" the bezier curves are.

@alarshi this is the change we talked about today, could you check if you can see any differences between your branch and this branch?

### DIFF
--- a/source/world_builder/objects/bezier_curve.cc
+++ b/source/world_builder/objects/bezier_curve.cc
@@ -358,7 +358,8 @@ namespace WorldBuilder
         }
       else
         {
-          for ( size_t cp_i = 0; cp_i < control_points.size(); ++cp_i)
+          const size_t max_cp_i = 1; //control_points.size();
+          for ( size_t cp_i = 0; cp_i < max_cp_i; ++cp_i)
             {
               const Point<2> &p1 = points[cp_i];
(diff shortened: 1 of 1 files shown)


alarshi:
@gassmoeller : I tried this branch with WB main and a lower resolution (4 + 2) and got different results (image below). There are several faults missing faults on this branch but I could not find a pattern for the difference.

image

gassmoeller:
Since it looks like this is not really working let me close this PR for now. We found another way to speed up the execution anyway.

Issue #493, opened 2023-06-01
Todo: release 0.5.0


MFraters:
- [X] run spell checker #494  and #495 
- [X] run clang-analyzer scan-build (https://clang-analyzer.llvm.org/command-line.html): No bugs found.
- [x] run clang tidy: #497
- [x] wait for reviews from ASPECT (https://github.com/geodynamics/aspect/pull/5164)
- [x] make new zenodo doi (before the release) to include in the new release. (#499)
- [x] increase version number (#499)
- [ ] make release

PR #497, opened 2023-06-01
Clang tidy fixes

MFraters:
- Made lot's of variables const. Seemed to prefer `type const` instead of `const type`, which I preferred, so I changed it.
- made some parameters into references
- It likes status member functions over const member functions. I don't really have an issue with this.
- reordered include files
- added comments to the end of namespaces with the name of that namespace.
- modernized some loops.
- added commented out parameter names for unused parameters
- some other small things like static_cast and std::move
- I found that in the composition uniform plugins, the declared `double composition` variable could be removed.  

These changes seem reasonable to me. Part of #493


MFraters:
Seems to also bring some performance improvements in some cases. @gassmoeller, what do you think?

source/world_builder/objects/surface.cc
```diff
@@ -114,6 +146,7 @@ namespace WorldBuilder
 
     }
 
+    <<<<<<< HEAD
```
gassmoeller, line 149: You want to remove this and the accompanying change in line 174

PR #501, opened 2023-06-06
post v0.5.0 release update.

MFraters:


MFraters:
The benchmarks fail because they use the updated version files with the old version in the binary. This should be fixed after merging.

Issue #502, opened 2023-06-25
warnings of version in ASPECT


tjhei:
```
/ssd/aspect-git/contrib/world_builder/include/delaunator-cpp/delaunator.hpp:348:16: warning: variable 'hull_size' set but not used [-Wunused-but-set-variable]
/ssd/aspect-git/contrib/world_builder/include/world_builder/world.h:301:12: warning: private field 'limit_debug_consistency_checks' is not used [-Wunused-private-field]
/ssd/aspect-git/contrib/world_builder/source/world_builder/objects/bezier_curve.cc:328:38: warning: unused variable 'second_derivative_point' [-Wunused-variable]
/ssd/aspect-git/contrib/world_builder/source/world_builder/objects/bezier_curve.cc:508:38: warning: unused variable 'second_derivative_point' [-Wunused-variable]
/ssd/aspect-git/contrib/world_builder/source/world_builder/objects/bezier_curve.cc:84:26: warning: unused variable 'P2P1' [-Wunused-variable]
/ssd/aspect-git/contrib/world_builder/source/world_builder/utilities.cc:610:30: warning: unused variable 'reference_normal' [-Wunused-variable]
```
This is with clang-15.

MFraters:
fixed by #503

PR #503, opened 2023-07-09
Fix warning reported on macs.

MFraters:


MFraters:
fixes #502 which was reported by @tjhei

Issue #199, opened 2020-07-08
Improve error output for wrong input files


MFraters:
The errors provided by the JSON parser are not very readable. We might be able to catch some of the common problems ourself and provide a more user friendly error to the users, or improve the parser output itself.

PR #511, opened 2023-08-25
Move the coordinates check outside of rapidjson into parse asserts.

MFraters:
This is work to prepare for the contours/slab2 integration issue (#429).

Will have to think a bit more about the post-processing of the error output data.


MFraters:
This is also I think a good first step towards #199.

PR #517, opened 2023-10-13
Fix date for v0.5.0 in citation

ljhwang:
Fixes the year for the citation to the software for v0.5.0 to 2023.

Also changes the day of the month to align with the citation on Zenodo.

This may not change everywhere where the citation is incorrect. Only the places I noticed  :)


MFraters (approved):
Great, thanks for the fix!

Issue #518, opened 2023-10-20
Discontinuous Spreading Center in Spherical Models When Passing 180 Lon


danieldouglas92:
When trying to write a .wb file that contains a mid-ocean ridge that crosses through 180 degree longitude, the ridge truncates at 180. 

I attached the output of an example world builder file that shows how the ridge gets stopped at 180 lon. It shows the temperature at the surface of the model that has dimensions -10 to 10 degrees latitude, and 170 to 190 degrees longitude. The mid ocean ridge should span go from 170 to 190 degrees longitude along the equator, but instead it stops in the center.

![truncated_ridge](https://github.com/GeodynamicWorldBuilder/WorldBuilder/assets/55814099/9e32c18d-3145-45db-8e5c-1c1318230a59)

[MOR_grid.txt](https://github.com/GeodynamicWorldBuilder/WorldBuilder/files/13056839/MOR_grid.txt)
[MOR_wb.txt](https://github.com/GeodynamicWorldBuilder/WorldBuilder/files/13056841/MOR_wb.txt)

MFraters:
Thanks for reporting this issue, I will take a look at it.

MFraters:
For reference, my initial guess is that this is related to, and probably needs a similar fix, as issue #338

danieldouglas92:
I tried messing around with the way I define the longitude and and I'm still getting discontinuity across 180.

Here is an example where the ridge coordinates are `[[[-191, 0], [-181, 0], [-179, 0], [-169, 0]]]`:

![image](https://github.com/GeodynamicWorldBuilder/WorldBuilder/assets/55814099/2d10a685-8fa6-495a-bbf5-4d0d58692880)


And here is an example where the Ridge coordinates are `[[[169, 0], [179, 0], [-179, 0], [-169, 0]]]`:

![image](https://github.com/GeodynamicWorldBuilder/WorldBuilder/assets/55814099/8b012cb9-b941-406d-adab-bdf602042a73)

danieldouglas92:
As a bypass to this issue I thought that I would try just shifting my entire region so that I wouldn't have to deal with crossing the prime meridian. This fixed the issue of the ridge being discontinuous but it seems to have introduced another issue with the temperature being incorrect at depth. 

This is the temperature at the surface showing a continuous ridge:
[...]

PR #520, opened 2023-11-21
Add tests and fix for spherical spreading centers

danieldouglas92:
Addresses Issue #518


MFraters (approved):
This looks good to me. Can you also add the fix for the mass conserving temperature model we discussed and update the test results?

MFraters (commented):
almost there, I think you forgot to add the .wb and .dat files for the test you made

Issue #522, opened 2023-12-04
Linear constant dip faults create circular features at depth


danieldouglas92:
When trying to create a constant dipping linear fault feature, one of the points on the fault turns into a radial feature at depth. I attached screenshots of the linear fault at the surface and at depth, showing this behaviour. I also attached the grid file and worldbuilder file that I used. 

![image](https://github.com/GeodynamicWorldBuilder/WorldBuilder/assets/55814099/50c75fdd-34da-470a-b82e-1515b38c9669)

![image](https://github.com/GeodynamicWorldBuilder/WorldBuilder/assets/55814099/d17703e7-bb88-47cc-8435-f9a80d67ffee)

[grid.txt](https://github.com/GeodynamicWorldBuilder/WorldBuilder/files/13541573/grid.txt)
[fault_rings.txt](https://github.com/GeodynamicWorldBuilder/WorldBuilder/files/13541574/fault_rings.txt)

MFraters:
Thanks for reporting the issue. I have an idea what could be causing the issue, but I will need some more time to investigate it. It seems for the points on one side to include the end point in the distance calculation, which is causing this circular shape. I heard form you that you had found a case where it doesn't do this, can you also post that input file? Thanks!

MFraters:
@danieldouglas92, well finding the issue was easier than expected (see #523). I will now need to do quite a bit of testing. With AGU and the holidays coming up, I can't promise that it will be done before the new year, but can you give it a go as well to see if that pull request breaks any of your current models?

MFraters:
closed by #523

PR #523, opened 2023-12-05
Only limit the Bezier curve closest point at the end of the whole curve

MFraters:
Looking at issue #522, I realized that the end points of slabs and fault (i.e. bezier curves) are not handled properly. The issues stems form wanting to find the closest point to a line consisting of segments. It can happen that the closest point found is just off the segment. In the interior this may be fine (especially if the derivatives are non-continuous, although I should revisit this in the future again because with Bezier Curves they now should be), but it should not happen at the edges of the full line itself.

This part at the end of line  lines can be easily tested for, which is what this pull request does. I tested it with the provided example and this fixes the issue in #522.

This may break tests and I need to do some more testing to look if it breaks anything else it should not break.


MFraters:
After looking into a few changed test results, the provided test case and some of my more complex models, I think this pull request is ready to merge. 

@danieldouglas92, could you have a look at this pull request to see if 1. it resolves the issue in #522, and 2. your models still work.

danieldouglas92:
I just tested and it does fix the issue! Below are two screenshots with the same .wb and .grd file with and without the changes in this PR.

image

image

PR #543, opened 2024-02-13
Remove bundled libraries from doxygen

gassmoeller:
Currently all the bundled libraries are included in the doxygen documentation, which means it is hard to find the actual World Builder classes in the doxygen documentation (see https://codedocs.xyz/GeodynamicWorldBuilder/WorldBuilder/annotated.html). This change excludes all non WorldBuilder files from the doxygen documentation.


MFraters (approved):
Thanks!

PR #544, opened 2024-02-13
Update name of python package

gassmoeller:
It looks like sphinx_book_theme is actually sphinx-book-theme, see https://pypi.org/project/sphinx-book-theme/. Maybe it changed its name. Let's wait for the documentation tester before merging this to make sure the new version still works.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
-sphinx_book_theme
+sphinx-book-theme
 myst-parser
 sphinx-copybutton
 sphinx-design


gassmoeller:
Ok, looks like the documentation tester succeeded with the new name. I compared the build logs of readthedocs before/after and they look the same. Probably sphinx_book_theme was the old name, but is still supported, but sphinx-book-theme is the new name (since it is the only one listed on pypi).

This should be good to go once the other testers have finished.

PR #546, opened 2024-02-13
Add breathe bridge doxygen to sphinx

gassmoeller:
Use breathe (https://breathe.readthedocs.io/en/latest/) to include the doxygen documentation into the sphinx documentation. Currently this doesnt look very good, because it includes all doxygen documentation on a single page.


gassmoeller:
This is as good as I can make this right now. I can build the doxygen documentation into breathe, but there are doxygen warnings that are interpreted as errors by readthedocs (so it works locally, but not for the online documentation). If someone wants to work on the errors or improve the design feel free to take a stab, otherwise this PR is just for future reference.

PR #547, opened 2024-02-13
fix a cookbook

tjhei:
gwb-grid fails without this change

@MFraters should the tester catch this? where is the logic that runs the ``./cookbooks/``?


tjhei:
I found two places where this is used.

MFraters (approved):
Great, thanks for finding this. This is ready to merge.

PR #549, opened 2024-02-13
update interpolation argument documentation

tjhei:
also see  #547

### DIFF
--- a/source/world_builder/features/interface.cc
+++ b/source/world_builder/features/interface.cc
@@ -49,7 +49,6 @@ namespace WorldBuilder
 
         WBAssertThrow(false,
                       "You provided an interpolation type which is not supported: " << string
-                      << "The options are none, linear, monotone spline and continuous monotone spline. "
                       << "This may be due to all options besides continuous monotone spline have been "
                       << "removed since version 0.5. It is best to remove the interpolation variable "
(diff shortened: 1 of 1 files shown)


source/world_builder/features/interface.cc
```diff
@@ -89,8 +88,8 @@ namespace WorldBuilder
 
                 prm.declare_entry("interpolation",Types::String("global"),
                                   "What type of interpolation should be used to enforce the minimum points per "
-                                  "distance parameter. Options are global, none, linear, monotone spline and "
-                                  "continuous monotone spline interpolation. If this "
+                                  "distance parameter. Options are 'global' and "
+                                  "'continuous monotone spline' interpolation. If this "
```
MFraters, line 92: Could you add a note here that this parameter is deprecated?

MFraters (approved):
one small comment, otherwise good to go.

source/world_builder/features/interface.cc
tjhei: is this because the global one is also only "continuous monotone spline"?

MFraters: yes, and I probably want to get rid of this option before the 1.0 release.

tjhei:
updated.

Issue #536, opened 2024-02-12
Global Earthquake Model with the current main


alarshi:
We have been using the Global Earthquake Model (Pagani et al., 2018) in our local branch of worldbuilder because the main branch is too slow/does not work with this model. We want to figure out why this is the case and then make sure we can import this plate boundary model using the main branch.

alarshi:
@gassmoeller found that the issue is caused because for all newton iterations an output string stream is generated, which is useful when the solution for `closest_point_on_curve` is not found (`bezier_curve.cc`). We can address this in future such that the string stream is created only when the solution fails.

alarshi:
This is not an issue anymore after relaxing the tolerances for finding the solution.

PR #553, opened 2024-02-13
Add a note for differences in the release and debug mode.

alarshi:
@gassmoeller and I noticed that the difference in speeds in the main GWB and our local branch of GWB for our Global models is only in the `debug` mode and not in the `release` mode (#536), with a majority of time spent in the output stream operator. This is not an issue to fix, but maybe it would be useful to document this somewhere. This PR addresses this.


gassmoeller (approved):
Thanks, looks good!

PR #554, opened 2024-02-13
re-enabling 2D for chunk. not clear why this option was removed.

mibillen:
This option existed before May 2023. It is not clear why it was removed and the assert added. However, Magali and Becky both use this option in the previous version (0.5 to check input structure before running in Aspect), without any problems and can't change to the new version without it.  So, since we don't currently know what the problem is, I have removed the assert for now to allow the 2D option to be used with the chunk option and the grid input file.

### DIFF
--- a/source/gwb-grid/main.cc
+++ b/source/gwb-grid/main.cc
@@ -778,7 +778,7 @@ int main(int argc, char **argv)
           const double inner_radius = z_min;
           const double outer_radius = z_max;
 
-          WBAssertThrow(dim ==3, "2D is currently not supported for the chunk.");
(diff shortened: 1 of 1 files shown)


gassmoeller (commented):
Menno will have to decide on this PR (I think you discussed it with him), but I would suggest to delete the line completely if you are not sure why it was introduced in the first place. Leaving commented code is a recipe for slowly cluttering the whole code with untrustworthy and outdated information.

MFraters (approved):
Yes, I agree, please delete the line. Otherwise good to go. If we find an issue in the future, we can document directly what the issue is and/or directly try to fix it.

Issue #552, opened 2024-02-13
VS Code json autocompletion


tjhei:
VS Code understands "json with comments" files, but one needs to tell it to open ``.wb`` file as json files before coloring, etc. works.

One can also specify a schema to use for autocompletion:
```
{
  "version": "0.6",
  "$schema": "https://f.tjhei.info/gwb/gwb-0.6.schema.json",
}
```
Here I just uploaded a schema to one of my machines.

You can either specify it inside each file as above, or you can specify a default schema in your user settings:
```
    "json.schemas": [
        {
            "fileMatch": [
                "*.wb"
            ],
            "url": "https://f.tjhei.info/gwb/gwb-0.6.schema.json"
        }
    ]
```

Finally, we should also document to use jsonc by default for .wb files:
```
    "files.associations": {
        "*.wb": "jsonc"
    },
```

more details: https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings

tjhei:
... and now with documentation:
![image](https://github.com/GeodynamicWorldBuilder/WorldBuilder/assets/1531285/c1e77074-e1e7-434b-a562-96795de27c4b)

PR #556, opened 2024-02-13
json schema: do not output NaN

tjhei:
This trips up VSCode otherwise because ``NaN`` is not valid json. We could also remove the offending parameter:
https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/a01bf98bec4de35811e8ff11c463ce8197924a5b/source/world_builder/features/subducting_plate_models/temperature/plate_model.cc#L78

(this is the only one)

Thoughts?

part of #552


MFraters (approved):
Yes, this looks good to me.

PR #557, opened 2024-02-13
json schema: allow $schema

tjhei:
VSCode can parse embedded URLs to a schema file if there is a top level ``$schema`` value. Our schema needs to allow this.


MFraters (approved):
Great, thanks!

PR #560, opened 2024-02-13
json schema: use description

tjhei:
VSCode uses "description" to show documentation and not "documentation". This seems to be the correct convention, see
https://json-schema.org/learn/miscellaneous-examples for example.

part of #552


tjhei:
note, we need to merge #556 and #557 and then update the schema output before merging this.

@MFraters where is the "documentation" used? I did not get any errors.

PR #563, opened 2024-02-14
Add transform fault cookbook

jdannberg:
Cookbook for making a transform fault and creating an initial condition for ASPECT. 
This still has a WIP label because ideally we would wan to link to the corresponding ASPECT cookbook as well (which I haven't created yet).


MFraters (approved):
Looks great, thanks for writing this! The only comment I have is if you could split the lines a bit more, because now you have to scroll a lot to the side in the html documentation.

jdannberg:
Thank you! I addressed your comments and also added a changelog entry. 

I also made the corresponding ASPECT pull request, which can be found here: https://github.com/geodynamics/aspect/pull/5575.

jdannberg:
I fixed the changelog entry

gassmoeller (approved):

PR #564, opened 2024-02-14
Introduces the World Builder as a proper noun

ljhwang:
The main purpose is to introduce referencing the code as GWB or the World Builder. The change here is that World Builder is considered a proper noun and should be capitalized.

This also cleans up some grammar and punctuation.


MFraters (approved):
Thanks, looks good to me!

PR #565, opened 2024-02-14
Create plugin documentation

alarshi:
This PR adds documentation on how to write a new plugin in GWB. It also fixes small typos that I came across while working on the write up.

### DIFF
--- a/doc/sphinx/developer_manual/developing_for_the_GWB/creating_new_plugins.md
+++ b/doc/sphinx/developer_manual/developing_for_the_GWB/creating_new_plugins.md
@@ -2,6 +2,17 @@
 Creating new plugins
 ====================
 
-```{todo}
-Explain how to create a new plugin. If possible find one which already nearly does what you want and copy and change it, but also explain how to build one from sratch.
-```
+It is possible that the user wants to implement a different distribution of properties (temperature or composition) than the ones available in the GWB. This can be achieved by creating a new plugin with the steps outlined subsequently. We will discuss two ways to write a plugin: one is to base it on an existing plugin, and another is to write the plugin from scratch.
+
+#### Modify an existing plugin
+
+Identify an existing implementation that nearly does what you want and copy the contents into another file, say `myplugin.cc`, in the same directory. It is also recommended that you make a copy of the corresponding header file into `myplugin.h`. This allows other plugins to use the declarations in `myplugin`.
+
+As an example, assume that you want to implement faults in your world that have compositional value that varies according to a hat function. In this case, the closest existing functionality is the `smooth` fault composition that varies following a hyperbolic tangent function. You can then modify the implementation of the relevant function, i.e., `get_composition()`, such that it now returns a compositional value based on the hat function.
+
+#### Write a plugin from scratch
+
+Create your plugin file that contains the declarations and implementations of all the members of your class, which must be derived from the relevant `Interface` class. For example, if you are creating a new temperature computation for a `continental_plate_model`, then you would include the function declarations of `features/continental_plate_models/fault_models/interface.h`. This implies that you need to have the functions `declare_entries`, `parse_enteries`, and `get_temperature` and their defintions for your plugin. Similar to above, it is recommended that you split the function declarations into a header file, say `myplugin.h`.
+
+-----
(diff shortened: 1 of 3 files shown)


doc/sphinx/developer_manual/developing_for_the_GWB/creating_new_plugins.md
```diff
@@ -2,6 +2,17 @@
...
+
+As an example, assume that you want to implement faults in your world that have compositional value that varies according to a hat function. In this case, the closest existing functionality is the `smooth` fault composition that varies following a hyperbolic tangent function. You can then modify the implementation of the relevant function, i.e., `get_composition()`, such that it now returns a compositional value based on the hat function.
+
+#### Write a plugin from scratch
+
+Create your plugin file that contains the declarations and implementations of all the members of your class, which must be derived from the relevant `Interface` class. For example, if you are creating a new temperature computation for a `continental_plate_model`, then you would include the function declarations of `features/continental_plate_models/fault_models/interface.h`. This implies that you need to have the functions `declare_entries`, `parse_enteries`, and `get_temperature` and their defintions for your plugin. Similar to above, it is recommended that you split the function declarations into a header file, say `myplugin.h`.
```
gassmoeller, line 15: defintions -> definitions

doc/sphinx/developer_manual/developing_for_the_GWB/creating_new_plugins.md
```diff
@@ -2,6 +2,17 @@
...
+
+-----
+Register the plugin at the bottom of `myplugin.cc` in `WB_REGISTER_*` that instantiates the plugin, documents it, and makes it available to the parameter file handlers. Finally, compile the plugin, using `cmake .` and then `make` in the top directory.
```
gassmoeller, line 18: the "top" directory is not very clear. You probably mean the "build" directory? (which can be the top directory if you do an in source build, but could be anywhere else if you do an out of source build)

gassmoeller (commented):
Very nice, just two small comments.

MFraters (commented):
Great, thanks for writing this up! 

I think it might also be useful to add here that you will need to change the Class names in both the header and source files, and change the headerguard in the header file. Furthermore, I think it would be useful to add what information you need to put in the register function (e.g. ClassName, name used in the input file). Could you add something like this?

alarshi:
Thank you for the review! I addressed your comments in the recent push. Let me know if we need more details.

doc/sphinx/developer_manual/developing_for_the_GWB/creating_new_plugins.md
```diff
@@ -2,6 +2,17 @@
 Creating new plugins
 ====================
 
-```{todo}
-Explain how to create a new plugin. If possible find one which already nearly does what you want and copy and change it, but also explain how to build one from sratch.
-```
\ No newline at end of file
+It is possible that the user wants to implement a different distribution of properties (temperature or composition) than the ones available in the GWB. This can be achieved by creating a new plugin with the steps outlined subsequently. We will discuss two ways to write a plugin: one is to base it on an existing plugin, and another is to write the plugin from scratch.
+
+#### Modify an existing plugin
```
gassmoeller, line 7: You need to use ## to make readthedocs happy (you always start with # as the main heading and cannot reduce the heading level by more than one for each heading).

doc/sphinx/developer_manual/developing_for_the_GWB/creating_new_plugins.md
```diff
@@ -2,6 +2,17 @@
...
+
+Identify an existing implementation that nearly does what you want and copy the contents into another file, say `myplugin.cc`, in the same directory. It is also recommended that you make a copy of the corresponding header file into `myplugin.h`. This allows other plugins to use the declarations in `myplugin`. Then, rename the class in both the header and the source files to `Myplugin` (the class names are written in the Camel case by GWB convention) and modify the header guard accordingly.
+
+As an example, assume that you want to implement faults in your world that have compositional value that varies according to a hat function. In this case, the closest existing functionality is the `smooth` fault composition that varies following a hyperbolic tangent function. You can then rename the class to say Hat and modify the implementation of the relevant function, i.e., `get_composition()`, such that it now returns a compositional value based on the hat function.
+
+#### Write a plugin from scratch
```
gassmoeller, line 13: same as above

alarshi: Got it, thank you !

Issue #534, opened 2024-02-01
Mass conserving temperature: slab temperature with plate + slab reconstruction


lhy11009:
@mibillen @MFraters 
We are working towards applying the mass-conserving temperature to slab considering it's past converging rates, dip angles, trench motions, etc.
First, we need to validate the mass-conserving temperature profile by benchmarking it to 2-d kinematically driven subduction models.
We also need to consider the change in the ridge position of the plates, in the subduction converging rates, and the dip angles

Here are lists of tasks to complete:
- [ ] Benchmark of the Mass conserving temperature model with kinematically driven 2-d subduction models
- [ ] A consistent model with plate + slab reconstruction with movement with ridge and its spatially dependent spreading rate 
- [ ] A consistent model with variation in convergence rates and dipping angles (if that's possible) in geologic time
- [ ] Smoothing of the temperature profile at the coldest part of the slab (now there is a temperature gradient from the top and bottom parts of the slab).

MFraters:
It seem to me that part 2 is the same as #529, is that correct? Also #533 might be relevant for you.

Issue #533, opened 2024-01-25
Spatially Dependent Spreading Rates


danieldouglas92:
When defining a ridge, the spreading rate is assumed to be constant along the length of the entire feature. For some spreading centers, this is unrealistic (like the Lau spreading center), where there is a substantial decrease in spreading rate south along the ridge. To avoid having to manually break up a single ridge into multiple smaller ridges, each with it's own manually defined constant spreading rate, it would be smoother to implement a way to linearly change the spreading rate from v1 at ridge coordinate (x1, y1), to v2 at ridge coordinate (x2, y2).

The input for this in the temperature features could just expand the "spreading velocity" parameter into an array. This array would be of length N, where N is the number of user defined "ridge coordinates". 

`"ridge coordinates":[[[x1, y1], [x2, y2], ..., [xn, yn]]]`
`"spreading velocity":[v1, v2, ..., vn]`

And the spreading velocity would be linearly interpolated from `v1` to `v2` from `[x1, y1], [x2, y2]`, and then `v2` to `v3` from `[x2, y2], [x3, y3]` etc. If the user inputs a single value for `"spreading velocity"`, a constant value is used for the entire ridge feature.

MFraters:
Thanks for making the issue! I have been separately thinking about temporary varying spreading velocities. What you could do for now is just add an extra [[]] around the spreading velocities, so that in the future we could implement something like `"spreading velocity":[[0,[v1, v2, ..., vn]],[50e3,[v1, v2, ..., vn]]]`

This would then give us the following variants: 
1. no variation: `"spreading velocity":1e-2` or `"spreading velocity":[[[1e-2]]]`
2. temporal variation: `"spreading velocity":[[0,[1e-2]],[50e3,[5e-2]]]`
3. spatial variation:  `"spreading velocity":[[[1e-2,5e-2, 2.5e-2]]]`
4. temporal and spatial variation:  `"spreading velocity":[[0,[1e-2,5e-2, 2.5e-2]],[50e3,[2e-2,6e-2, 7.5e-2]]]`

What do you think? Will this become too complicated? Just to reterate, you would not have to implement case 2 and 4, but just design it in a way it would become possible to add it on later without breaking input files.

danieldouglas92:
I think that that sounds good! Also nice that it has a similar syntax as the variable overriding plate thickness

PR #579, opened 2024-02-14
Spatially Variable Spreading at Ridges

danieldouglas92:
Address Issue #533. Adds spatial variation for spreading centers in the half space temperature feature for oceanic plates. This requires a new parsing function that I called value_at_array that allows worldbuilder to extract an array of size N for a single value. This was done to prepare for new functionality that will be added to address #534. This will allow the user to specify the spreading rate at past times as well as specify spatial variations.


MFraters (approved):
I gave some comments in person. Otherwise looks good to me.

PR #585, opened 2024-02-14
gwb-grid: support for filtering cells

tjhei:
Add a new option --by-tag that enables filtering of cells inside gwb-grid. To actually do something useful, we need support inside GWB to provide tag information. So, right now, this option randomly selects some cells to be shown.

### DIFF
--- a/source/gwb-grid/main.cc
+++ b/source/gwb-grid/main.cc
@@ -33,6 +33,9 @@
 #include "world_builder/world.h"
 #include "world_builder/config.h"
 
+#include <algorithm>
+
+
 #include "vtu11/vtu11.hpp"
 #undef max
 #undef min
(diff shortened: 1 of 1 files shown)


tjhei:
This is ready for review @MFraters

MFraters (approved):
I have some questions about the structure, but that is probably because it is not actually using the tags yet. I think it is probably best to merge it as is, and then we can discuss those things when it is ready to do the actual filtering.

MFraters:
Or do you want to do everthing in this pull request, given that it is still marked as [wip]?

tjhei:
Let me make a few updates...

tjhei:
Okay, this is now ready to review and merge. I do not advertise the new option in ``--help`` yet, because it only filters tag -1 for now. I will follow up with additional PRs later. @MFraters

PR #587, opened 2024-02-14
gwb-grid: use threads by default

tjhei:
I did not benchmark if this is a good choice, but it is certainly faster than 1 thread. I happy to use something simpler (hardware_concurrency should be fine as well).


gassmoeller (approved):
I am fine with this, and the benchmarks seem to agree. I will let @MFraters take another look.

MFraters (approved):
Nice! The benchmarks definitely seem to like this. I also like that is show the default when using --help

PR #591, opened 2024-02-14
Create a function in Utilities.cc that calculates distance from Ridge + Spreading velocity

danieldouglas92:
Remove duplicate code in multiple places by creating a function in Utilities.cc that calculates the distance from a ridge and the spreading velocity. This is used by several different temperature plugins of various features. Currently, the spreading velocity is just a single, constant value, but this will be changed in a followup PR.


include/world_builder/utilities.h
```diff
@@ -426,6 +426,13 @@ namespace WorldBuilder
     */
     std::string
     read_and_distribute_file_content(const std::string &filename);
+
+
```
MFraters, line 430: Can you add documentation here?

MFraters (approved):
Looks good. Could you just add some documentation to the function?

PR #596, opened 2024-02-15
Update installation_methods.md

ljhwang:
Grammar and updates to the usage of World Builder.

Line 59: "The only thing you need to care for ..." is awkward but I do not have a suggested change.

Feel free to suggest some rewording.


MFraters (approved):
Maybe `take care of`. But it could also e repharsed as: "Note, when creating...".

Otherwise looks good to me.

Issue #532, opened 2024-01-25
[Tracking issue] Improve API and Developer documentation


MFraters:
The developer documentation is currently still mostly missing. This issue keeps track of the API and Developer manual which still need to be written. Please comment below if you are working on one of these pages, so no double work is done. Also don't forget to mention this issue in your pull request, so that it can be ticket off when it is merged.

**API manual**
- [ ] API design
- [ ] Available APIs
- [ ] API in different languages (FFI)

**Developer Manual**
- [ ] How to start developing and contributing
    - [ ] required and recommended tools
        - [ ] Git
        - [ ] Astyle
        - [ ] IDE
    - [ ] Contributing to the GWB
        - [ ] Helping others
        - [ ] Making a pull request
        - [ ] Pull request review process
            - [ ] What is continuous integration?
            - [ ] Testers
            - [ ] Code coverage
            - [ ] Benchmarking
            - [ ] Documentation testing
        - [ ] Pull request merging
- [ ] Contributing to the documentation
    - [ ] Important syntax
    - [ ] Documentation style guide
    - [ ] Documentation testing
    - [ ] Tips and tricks
    - [ ] The review process
- [ ] Contributing to the code
    - [ ] Code documentation and Doxygen 
    - [ ] Code conventions
    - [ ] Code Style Guide
    - [ ] Testing your code
        - [ ] Unit tests
        - [ ] integration testers
            - [ ] App tester
            - [ ] Visualization tester
            - [ ] Wrapper tester
    - [ ] Tips and tricks
    - [ ] Review process
- [ ] Developing for the GWB
    - [ ] Code structure
    - [ ] Creating new plugins
    - [ ] Changing or adding code
    - [ ] Creating a new plugin system
    - [ ] Creating new tools

alarshi:
I am working on `Creating new plugins`

PR #598, opened 2024-02-15
Add developer documentation for plugin systems.

alarshi:
This PR adds documentation for implementing new plugin system. #532


MFraters (commented):
Thanks for writing this up! One thing they would also need to add is a .cc/.h file in the features folder for the feature (for example https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/main/source/world_builder/features/continental_plate.cc). Can you add that?

Otherwise looks good to go.

alarshi:
Oh yes, I completely forgot to add that. Thank you for the review!

MFraters (approved):
Thanks for writing this!

PR #604, opened 2024-02-15
Only make html manual with "make manual"

gassmoeller:
This simplifies the `make manual` target to only build the html manual. In fact it just calls the `make manual_html` target to keep the CMakeLists.txt as simple as possible. I thought it is useful to keep the `manual_html` target as a contrast to the `manual_pdf` target that still exists.


MFraters (approved):
Thanks!

PR #620, opened 2024-02-16
add a new Plume feature

jdannberg:
I added a plume as a new feature, which allows it to make a cylinder with different elliptic cross sections, allowing it to specify their depth, eccentricity, semi-major axes and angles. 
Note that I just copied the temperature/composition/grains plugins from the continental plate, so let me know in case I need to change anything in there. I tested that this work s locally in 3d Cartesian and spherical model. 

Things I still want to do
- [x] Write a cookbook page to document the feature
- [ ] add an example to the basic starter tutorial
- [x] make some tests
- [x] test this in 2d
- [x] test this for the grains
- [ ] add a Gaussian interpolation for the temperature
- [x] make a spherical outline between the min depth and the first cross section, so that the tip of the plume will always be a sphere
- [x] add a changelog entry

Note that in spherical coordinates, this currently interpolates in lat/lon, and also computes the outline of the ellipse that way. This is something we might want to fix eventually, but I spent a few hours on this and it was more complicated that I had initially thought.


source/world_builder/features/plume.cc
```diff
@@ -0,0 +1,346 @@
...
+      prm.declare_entry("cross section depths", Types::Array(Types::Double(0)),
+                        "The depths of the elliptic cross section of the plume. Units: m.");
+      prm.declare_entry("semi-major axis", Types::Array(Types::Double(100.e3)),
+                        "The lengths of the semi-major axes of the elliptic cross sections of the plume. "
+                        "In spherical coordinates, this is in degrees, otherwise in meters.");
+      prm.declare_entry("eccentricity", Types::Array(Types::Double(0)),
+                        "The eccentricities of the cross sections.");
+      prm.declare_entry("rotation angles", Types::Array(Types::Double(0)),
+                        "The directions that the semi-major axis of the elliptic cross-sections "
+                        "are pointing to, as an angle from geographic North in degrees. "
```
MFraters, line 79: What does north mean in a Cartesian box?

source/world_builder/features/plume.cc
```diff
@@ -0,0 +1,346 @@
...
+      WBAssert(rotation_angles.size() == coordinates.size(),
+               "The rotation angles array needs to have the same number of entries as there are coordinates. At the moment there are: "
+               << rotation_angles.size()
+               << " rotation angle entries but "
+               << coordinates.size()
+               << " coordinates!");
+
+
+      // Convert degrees to radians, convert from geographical to mathematical
+      // TODO: convert semi_major_axis_lengths as well for spherical coordinates
```
MFraters, line 146: Is this todo still valid?

source/world_builder/features/plume.cc
```diff
@@ -0,0 +1,346 @@
...
+    Plume::properties(const Point<3> &position_in_cartesian_coordinates,
+                      const Objects::NaturalCoordinate &position_in_natural_coordinates,
+                      const double depth,
+                      const std::vector<std::array<unsigned int,3>> &properties,
+                      const double gravity_norm,
+                      const std::vector<size_t> &entry_in_output,
+                      std::vector<double> &output) const
+    {
+      // Figure out if the point is within the plume
+      auto upper = std::upper_bound(depths.begin(), depths.end(), depth);
```
MFraters, line 217: I think you need to add `#include <algorithm>`

source/world_builder/features/plume.cc
```diff
@@ -0,0 +1,346 @@
...
+                                                              depth,
+                                                              properties[i_property][1],
+                                                              grains,
+                                                              min_depth_local,
+                                                              max_depth_local);
+
+                          }
+                        grains.unroll_into(output,entry_in_output[i_property]);
+                      }
+                      break;
```
MFraters, line 334: It now also needs a case 4 and a tag. See 
https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/dfbb30bbb8164ea46f52a8f787ed964bcc8cde7d/source/world_builder/features/continental_plate.cc#L234-L244

and 

https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/dfbb30bbb8164ea46f52a8f787ed964bcc8cde7d/source/world_builder/features/continental_plate.cc#L89-L94

MFraters (commented):
The code looks good to me. Some small comments.

jdannberg:
I addressed your comments, and I also had to rebase to get the "tags" feature from the interface.

jdannberg:
I have also added two tests:

This one is for Cartesian geometry (high resolution above, actual resolution in the test below):
test1
test1_lowres

And this one is for spherical geometry (high resolution above, actual resolution in the test below):
test_spherical
test_spherical_low_res

jdannberg:
To document how the plume tip looks like:

ice_cream

MFraters (approved):

PR #628, opened 2024-02-16
Update point_features.md

ljhwang:
More editing for clarity.

Is the plume feature being implemented as a point feature?

Please feel free to edit.


MFraters (approved):
Thanks, when #620 is merged we should edit this again.

Issue #456, opened 2022-12-27
Follow up on nterface for the distance to a feature's plane


MFraters:
Follow up on #453. 

- [ ] implement it for faults as well
- [ ] make available through the different interfaces

PR #634, opened 2024-02-16
Add distance from plane in fault.

alarshi:
This PR addresses #456, i.e., add distance to the fault surface.


MFraters (approved):
I was wondering why the results where so different compared to the subducting plate one, since the names of the .wb files are so similar. It turns out that the contents are actually more different then I expected, so the different results make sense.

PR #635, opened 2024-02-16
doc: feature list

tjhei:
- put area/line/point features on a single page (pages were very short)
- include a list of all current features


MFraters (commented):
I think the one page is indeed better for this, but I think you used the text from before the edits Lorraine made. Can you update them?

MFraters (approved):

ljhwang:
I think I am late to the party. Is there something I need to do still?

I would also consider changing the headings from # to ##

PR #636, opened 2024-02-16
Update citing.md

ljhwang:
@tjhei 

Please check to see if this resolves your previous comments.


tjhei (approved):
Great. Did you see how it looks like in the final version?
https://gwb--636.org.readthedocs.build/en/636/user_manual/introduction/citing.html#data-availability

Maybe it makes sense to break the lines to make it more readable?

Issue #526, opened 2024-01-23
[Tracking issue] gwb-grid improvements


MFraters:
The following are subprojects to improve gwb-grid:
1. **Improve documentation of gwb-grid**: Currently there is a short description in the  manual here: https://gwb.readthedocs.io/en/latest/user_manual/how_to_use_the_applications/gwb-grid_app.html which can be expanded, but this page: https://gwb.readthedocs.io/en/latest/GWB_parameter_listings/world_builder_file/index.html should also be filled or removed.
2. **Improve testing of gwb-grid grids**: Currently only `cartesian` and `chunk` are tested, but there is also a `sphere` and an `anullus`. 
3. **The gwb-grid should be added to the coverage result**: maybe as a separate category, or just lump it in with the rest
4. **Better parallelization**: gwb-grid currently only has a very simple parallezation scheme. Some kind of job system would be better. (@tjhei?)
5. **Add vti output support**: this is useful for online visualization in point 7 (@tjhei)
6. **Add netcdf output support** (@tjhei)
7. **Implement online visualization**: through glance (https://pvw.kitware.com/glance/) (@tjhei)

tjhei:
Conversation about better parallelization:
- simple scheduler code based on c++ 11: https://blog.andreiavram.ro/job-scheduler-cpp/
- taskflow https://taskflow.github.io/taskflow/index.html#SupportedCompilers (header-only, but big and requires c++17)

tjhei:
Do you have an example where the current parallel scheduling is not ideal? I could not reproduce based on your description.

MFraters:
It mostly shows up at larger models. For example, if you double the resolution of in `cookbooks/3d_cartesian_curved_subduction/3d_cartesian_curved_subduction.grid` and then run ./bin/gwb-grid  -j 6 ../cookbooks/3d_cartesian_curved_subduction/3d_cartesian_curved_subduction.wb `../cookbooks/3d_cartesian_curved_subduction/3d_cartesian_curved_subduction.grid` (in release mode), you can see that the first few are done very quickly. The higher resolution, the more the problem creeps up. Doubling the resolution again will make it very clear.

But I noticed that when doing the second doubling, the fact that the creating the grid and the writing of the paraview file are not parallel, is costing way more time at that point.

PR #639, opened 2024-02-16
gwb-grid: by-tag and --filter

tjhei:
gwb-grid now has two new options:

--filtered: This produces an additional <myfile>.filtered.vtu that removes background cells and cells tagged as "mantle layer"

--by-tag: This additionally produces a sequence of files called <myfile>.N.vtu where N is an integer id of a "tag". Each file only contains the cells containing features with this tag.

part of #526


MFraters (approved):
We should make many of these things in the world private, but we can do that in a different pull request.

PR #640, opened 2024-02-16
Improve documentation of the fault feature code.

gassmoeller:
The only functional change is line 512 where previously we would not consider a point inside the fault, if its distance to the fault was exactly 0 (this looks like a bug, but hopefully wouldnt happen too often).


tjhei (approved):

tjhei:
maybe a changelog entry for the fix?

gassmoeller:
Changelog added.

Issue #535, opened 2024-02-09
Create cookbook for 2D subduction model using mass conserving temperature 


mibillen:
Create cookbooks to guide new users through constructing a 2D subduction model using the mass-conserving temperature structure and crustal composition layers.  Need two cases: one for a "generic" slab shape in cartesian coordinates using the plate model for temperature and no adiabatic temperature gradient, and a second case for a slab shape based on slab2.0, observed spreading rates and ridge location in spherical coordinates using a half-space temperature model and an adiabatic temperature gradient.  For both cases we will include a ridge  at the box boundary and free-slip BCs. These two cases will cover most of the different issues that a new user will face in creating the parameter file.  

Case 1 - To Do
- Define model space, trench and ridge locations, and subducting plate speed. 
- No back ground adiabatic gradient - explain how this is set or not set. 
- Define the temperature (plate model) and compositions for the surface plates (one continent, on oceanic)
- Define slab shape using multiple segments.
- Add the composition layers for the slab: use this step to define the parameters that control the slab thickness, and directions (above and below slab surface)
- Choose plate model for mass conserving temperature for the slab : explain other parameters that control how slab temperature is smoothed into the background including at the slab tip
- Make illustrative figures at different steps to explain each step. 

Case 2 - South America - Nazca Plate 
- Define model space, trench and ridge locations, and subducting plate speed. 
- Add ground adiabatic gradient - explain how this is set or not set. 
- Define the temperature (half-space model) and compositions for the surface plates (both oceanic)
- Define slab shape using point and angle extracted from Slab 2.0 (need to makepygmt,  juptyer notebook available for this part)
- Add the composition layers for the slab: use this step to define the parameters that control the slab thickness, and directions (above and below slab surface)
- Choose half space model for mass conserving temperature for the slab : explain other parameters that control how slab temperature is smoothed into the background including at the slab tip and wedge corner.
- Make illustrative figures at different steps to explain each step.

mibillen:
@MFraters - Can you apply the hackathon label to this issue. I don't think I have permissions to add labels.

MFraters:
Thanks for writing this up,it would be great to have these cookbooks! I added the label.

MFraters:
@mibillen I think this can be closed now, or did you have further plans?

mibillen:
yes it can be closed

PR #648, opened 2024-02-17
Add a simple subduction 2d chunk model cookbook

mibillen:
This pull request address issue #535, case 2 although I decided to keep this cook book simpler, so it does not include a location specific model of points from slab 2.0.  It does however, several small but tricky things about making 2D models in the spherical/chunk geometry.


gassmoeller (commented):
Very nice, thanks for the detailed explanation (my way of learning the world builder is to review everyone's pull requests :-)).
One question about the paraview state files, everything else is just minor typos.

MFraters:
@mibillen I think you forgot the check in `/cookbooks/simple_subduction_2d_chunk/simple_subduction_2d_chunk.grid`. The tester need that to test the world builder file.

mibillen:
I've added the missing file ... the needed file was there, but it had a different name. I committed and then "squashed" :-)

MFraters (approved):
Great, cookbook, thanks for writing this! It i very detailed and leads the user through your thought process in setting up this model. I also like that you put some try yourself exercises at the end :)

I found a few more of Rene's suggestions which for some reason didn't show up in the conversations page, but did in the files changed. They looked uncontroversial to me, so I just merged them.

PR #650, opened 2024-02-17
add features JSON snippets

tjhei:
I am open to improvements to the content of the snippets, but this is quite useful in its current state.


tjhei:
maybe merge #650 first and I will update the schema...

tjhei:
rebased.

tjhei:
updated to fix documentation build.

source/world_builder/features/continental_plate.cc
```diff
@@ -56,12 +56,44 @@ namespace WorldBuilder
...
+        "temperature models":[{"model":"uniform", "temperature":600.0}],
+        "composition models":[{"model":"uniform", "compositions": [0], "fractions":[1.0]}]
+       }
+       */
+
+      Pointer((path + "/body").c_str()).Set(declarations,"object");
+      Pointer((path + "/body/model").c_str()).Set(declarations,"continental plate");
+      Pointer((path + "/body/name").c_str()).Set(declarations,"${1:My Plate}");
+      Pointer((path + "/body/coordinates").c_str()).Create(declarations).SetArray();
+      //Pointer((path + "/body/coordinates/0").c_str()).Create(declarations).SetArray();
```
MFraters, line 84: Either remove this, or enable it. I am fine either way. Or you could put a 0.0,0.0 coordinate in there.

source/world_builder/features/fault.cc
```diff
@@ -51,6 +51,39 @@ namespace WorldBuilder
...
+        "segments": [],
+        "sections": [],
+        "temperature models":[{"model":"uniform", "temperature":600.0}],
+        "composition models":[{"model":"uniform", "compositions": [0], "fractions":[1.0]}]
+       }
+       */
+
+      Pointer((path + "/body").c_str()).Set(declarations,"object");
+      Pointer((path + "/body/model").c_str()).Set(declarations,"fault");
+      Pointer((path + "/body/name").c_str()).Set(declarations,"${1:Menno's Fault}");
```
MFraters, line 78: I am not contradicting the default name here ;) Still, it think it might be better to align all the default names. I like the `My ...` version, because it indicates that you can write something there, more then the lower case `mantle layer` for example.

source/world_builder/features/continental_plate.cc
tjhei: Yeah, one problem is that if you add [] or [0,0], rapidjson ends up putting every single bracket on a single line, which then literally appears when you insert the template. I see the following options:
1. Leave coordinates empty as it is right now.
2. Modify rapidjson to influence the formatting.
3. Insert the snippets manually into the json schema with a script instead of directly in code. That way you can control precisely what formatting you would like.

Preferences?

MFraters: ah, sorry for the late reply. I somehow missed this. Let's leave it as is now. Then we can get it in as part of 0.6 and see from there.

tjhei:
updated.

MFraters (approved):
This looks great. Also a nice cleanup of constructor of Paraemeters. This looks ready to merge to me, can you confirm that, or do you want to play around with it some more?

tjhei:
Good to go from my side.

Issue #531, opened 2024-01-24
bounded random compositional plugins


MFraters:
Random compositional fields can be useful in certain cases, such as to initialize initial strain. It would be useful to add 
1. a plugin which picks a random number (between user defined bounds) and assigns that as a fraction at that location
2. a plugin which computes a more smooth random field, like a [perlin noise](https://en.wikipedia.org/wiki/Perlin_noise) to assign. The user could then tweak the parameters of the noise to their liking.

alarshi:
I will try to address this.

MFraters:
I think this is resolved by #651. Feel free to reopen is not. Thanks for working on this @alarshi!

PR #651, opened 2024-02-17
Add random compositional fields

alarshi:
This PR addresses part of issue #531 to get random compositional values. Here is the output of this plugin with a continental plate:

image
This is the result for user specified bounds in the first compositional field: [5, 10]

image
This is the result from using the default `random` model.


source/world_builder/features/continental_plate_models/composition/random.cc
```diff
@@ -0,0 +1,138 @@
...
+            {
+              const double min_depth_local = min_depth_surface.constant_value ? min_depth : min_depth_surface.local_value(position_in_natural_coordinates.get_surface_point()).interpolated_value;
+              const double max_depth_local = max_depth_surface.constant_value ? max_depth : max_depth_surface.local_value(position_in_natural_coordinates.get_surface_point()).interpolated_value;
+              if (depth <= max_depth_local &&  depth >= min_depth_local)
+                {
+                  for (unsigned int i =0; i < compositions.size(); ++i)
+                    {
+                      if (compositions[i] == composition_number)
+                        {
+                          const double composition_ = ((double)rand()) / RAND_MAX * (bound_values[1] - bound_values[0]) + bound_values[0];
```
MFraters, line 121: This works, but the following would be more in line with how the rest of the code generates random numbers.

```suggestion
                          std::uniform_real_distribution<> dist(min_value,max_value);
                          const double composition_ = dist(world->get_random_number_engine());
```

source/world_builder/features/continental_plate_models/composition/random.cc
```diff
@@ -0,0 +1,138 @@
...
+          prm.declare_entry("min depth", Types::OneOf(Types::Double(0),Types::Array(Types::ValueAtPoints(0., 2.))),
+                            "The depth in meters from which the composition of this feature is present.");
+
+          prm.declare_entry("max depth", Types::OneOf(Types::Double(std::numeric_limits<double>::max()),Types::Array(Types::ValueAtPoints(std::numeric_limits<double>::max(), 2.))),
+                            "The depth in meters to which the composition of this feature is present.");
+
+          prm.declare_entry("compositions", Types::Array(Types::UnsignedInt(),0),
+                            "A list with the labels of the composition which are present there.");
+
+          prm.declare_entry("bound values", Types::Array(Types::Double(1), (0, 1)),
```
MFraters, line 77: Can you turn this into a `Max value` and a `Min value` or something similar? I think it is better to have them as separate entries.

source/world_builder/features/continental_plate_models/composition/random.cc
```diff
@@ -0,0 +1,138 @@
...
+        void
+        Random::parse_entries(Parameters &prm, const std::vector<Point<2>> &coordinates)
+        {
+          min_depth_surface = Objects::Surface(prm.get("min depth",coordinates));
+          min_depth = min_depth_surface.minimum;
+          max_depth_surface = Objects::Surface(prm.get("max depth",coordinates));
+          max_depth = max_depth_surface.maximum;
+
+          compositions = prm.get_vector<unsigned int>("compositions");
+          bound_values = prm.get_vector<double>("bound values");
```
MFraters, line 97: Can you split it here as well?

MFraters (commented):
Looks great, thanks for making this! I have a few small comments, but otherwise I think it is good to go.

alarshi:
Thank you for the review! I addressed the comments in the latest commit. 
I was also wondering if the test case makes sense or not. Since, this is a random distribution, the output would be different every time we run it or do I not understand this correctly?

MFraters:
> I was also wondering if the test case makes sense or not. Since, this is a random distribution, the output would be different every time we run it or do I not understand this correctly?

The random numbers are depended on the seed. As long as the seed is the same, the same sequence of random numbers will be generated. If you update your test results, the tester should pass.

alarshi:
Got it! Thanks for the explanation.

MFraters:
hmm, strange. Did you rebase on the newest master when updating the test results?

alarshi:
I rebased the branch but the tests are still failing. I don't understand why. :(

alarshi:
I checked that each run is gives me a different result, which would explain why the test is failing.

tjhei:
The problem is that gwb-grid now uses multiple threads by default. This means that the points will not be evaluated in the same, deterministic order. I am not even sure that the random engine is thread-safe.
Making tests deterministic is not too difficult, by limiting all (or at least this) test to run with -j 1. If you want it to be deterministic for >1 thread, we would need to duplicate each random engine into a thread-local variable. This would make it deterministic as long as you run with the same number of threads but it would still depend on how many threads you run with. I am not sure if this is good enough...

MFraters:
ah, that make sense. I was confused, because we currently already use random number generation in the code, but this is for the grains, which is only implemented for the for gwb-dat, which doesn't use threads.

MFraters:
In any case, we should probably fix the number of threads for the tester.

MFraters:
hmm, getting a thread rank is not as easy as getting a mpi rank. The thread id seems to be an unpredictable number/hash (I think). We do actually have a unique rank number when creating them, but we would need to pass those id's manually somehow.

Looking at it, the easiest solution may be to add a constructor parameter to the world so that you can pass a thread rank. That rank is then use in some way to set the seed of the world. In gwb-grid, we then just create a separate world for each thread, and in the parallel for loop we call the properties function on the appropriate world for each thread. This would make it deterministic again and would require minimal changes I think.

MFraters:
@alarshi #667 is merged now, so if you rebase you should be able to fix this problem (at least updating the test results on my systems seems to work now). Also, it looks like you will need to include a header (I think the <random> header?) to make the non-unity build work.

MFraters (approved):
I am not sure why it didn't work for you, but it seems to work for me :s

PR #655, opened 2024-02-17
Reduce unnecessary code and fix a typo

danieldouglas92:


MFraters (approved):
You are removing a *1? Hope the tests are not breaking because of that ;)

PR #657, opened 2024-02-18
add a Gaussian temperature interpolation for the plume feature

jdannberg:
I am here adding two images of what the test cases would look like with a grid file (but I thought they'd be faster to run with gwd-dat):

Cartesian:
plume_gaussian

Spherical:
plume_spherical


MFraters:
The pictures look really cool! I will do my best to do a review in the coming week.

source/world_builder/features/plume_models/temperature/gaussian.cc
```diff
@@ -0,0 +1,166 @@
...
+                            "the plume as given by the plume feature) can be changed with depth. "
+                            "Temperature is always interpolated in a horizonzal/radial plane, except for the plume "
+                            "head: If the first depth of the plume centerline and the minimum depth of the plume "
+                            "feature are different, an ellipsoidal plume head is created in this depth range. "
+                            "Within this plume head, temperature is interpolated radially, i.e., depending on the "
+                            "distance from the center of the ellipsoid.");
+
+          prm.declare_entry("depths", Types::Array(Types::Double(0)),
+                            "The temperature at the center of this feature in degree Kelvin."
+                            "If the value is below zero, the an adiabatic temperature is used.");
```
MFraters, line 72: I don't think this description is correct.

source/world_builder/features/plume_models/temperature/gaussian.cc
```diff
@@ -0,0 +1,166 @@
...
+                            "The sigma (standard deviation) of the Gaussian function used to compute the "
+                            "temperature distribution within the plume. This sigma is non-dimensinal, i.e. "
+                            "it is defined relative to the distance between the plume center and margin as "
+                            "defined by the plume feature. Choosing a sigma of 1 therefore means that the "
+                            "temperature at the plume margin is set to a fraction of 1/sqrt(e) (approx. 0.61) "
+                            "of the centerline temperature. To achieve a smoother transition between the "
+                            "plume temperature and the outside temperature a smaller values has to be chosen "
+                            "for the gaussian sigmas.");
+
+          // TODO: assert that the three have the same length
```
MFraters, line 86: This todo belongs in parse_entries, and it looks like it is already done. I think you can just remove this.

MFraters (approved):
Looks good, just two minor comments. Can you also update the test reference result? Otherwise good to go.

MFraters (approved):

MFraters:
hmm, I think you still need to include [include/world_builder/features/feature_utilities.h](https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/d6df754e5e5d0b14b14d080ff1ac06e84c1ceecb/include/world_builder/features/feature_utilities.h#L35)

jdannberg:
I also looked at the test output of the changed test (in higher resolution):

comparison

The two main differences are that the very tip is now included as well in the sphere at the top (which is a rounding off question somewhere somehow I suspect), the other one is that the base of the plume tail is drawn to the max depth of the plume feature, even if it's below the deepest cross section given (which is a clear bugfix, because otherwise the max depth would be useless).

MFraters:
Sounds good to me. I think it should be good to go as soon as you include [include/world_builder/features/feature_utilities.h](https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/d6df754e5e5d0b14b14d080ff1ac06e84c1ceecb/include/world_builder/features/feature_utilities.h#L35)

jdannberg:
Okay, after several tries I think I now include all the things I need to include.

MFraters:
Can you still add a changelog entry in a follow up pull request?

jdannberg:
Sure. I thought this would just be a part of the plume feature. Do you want me to add it there or make a separate entry?

MFraters:
hmm, it is adding a new temperature model to the plume feature, so in that sense I think it is a notable change. I will leave it up to you.

PR #652, opened 2024-02-17
Fix ValueAtPoints data type

tjhei:
max_values_in_array should be an integer not a double.

Also update schema files.

PR #659, opened 2024-02-18
Apply spline to the mass conserving temperature

lhy11009:
In this PR, I applied a spline to the mass-conserving temperature. Previously, there was a high-order discontinuity at the coldest part of the slab. This additional treatment would generate a continuous profile by the spline operation.

To show the effect, below is the visualization of a slab temperature (from the new test mass_conserving_slab_temperature_spline.wb).
The line features show 2 profiles across the slab.
mass_conserving_spline_new_T

One free variable is the number of points on the spline.
To maintain the minimum temperature in the slab, one point is always pinned on the coldest point of the slab.
In addition to that, an equal number of points are chosen on both sides (default is 5).
A profile at 500 km depth (the deeper one in the first plot) is chosen to show the temperature profile from different numbers of spline points.

tmass_conserving_slab_temperature_spline

n = 2 there is a strong smoothing effect on the slab's internal temperature.

n = 5, the spline only affects the center of the slab as designated. 
As other parts of the profile are already continuous, the spline would follow the analytic profile.

n = 10, there is little effect on the temperature profile.


lhy11009:
@MFraters I see there are file differences not from the modification I committed in "File Changed". I only modified the source/world_builder/features/subducting_plate_models/temperature/mass_conserving.cc and the include/world_builder/features/subducting_plate_models/temperature/mass_conserving.h

MFraters:
Thanks for this pull request! I think it would be good for @mibillen to take a first look at it, given that she is more familiar with the mass conserving temperature model

> @MFraters I see there are file differences not from the modification I committed in "File Changed". I only modified the source/world_builder/features/subducting_plate_models/temperature/mass_conserving.cc and the include/world_builder/features/subducting_plate_models/temperature/mass_conserving.h

I assume you also added the test files? As for the doc/world_builder_declarations files, for now don't worry about them. You can remove them from you commits for now if you want. They will be mostly cleaned up in #652, althoug, because you are adding an option, these files will change in the end.

source/world_builder/features/subducting_plate_models/temperature/mass_conserving.cc
```diff
@@ -169,6 +170,13 @@ namespace WorldBuilder
           prm.declare_entry("reference model name",  Types::String("half space model"),
                             "The type of thermal model to use in the mass conserving model of slab temperature. "
                             "Options are half space model and plate model");
+
+          prm.declare_entry("apply spline",  Types::Bool(false),
+                            "Whether a spline should be applied on the mass conserving model.");
```
mibillen, line 175: change  "Whether a spline should be applied on the mass conserving model."
to "Choose to apply a spline to smooth temperature where the two analytical equations meet at the minimum temperature."

source/world_builder/features/subducting_plate_models/temperature/mass_conserving.cc
```diff
@@ -169,6 +170,13 @@ namespace WorldBuilder
...
+
+          prm.declare_entry("number of points in spline", Types::Int(5),
+                            "The number of points in the spline");
```
mibillen, line 178: Change "The number of points in the spline"
to "The number of points on each side of the spline centered at the point  where the two analytic equations meet. A default value of 5, will lead to a total of 11 points in the spline."

source/world_builder/features/subducting_plate_models/temperature/mass_conserving.cc
```diff
@@ -473,8 +484,35 @@ namespace WorldBuilder
...
 
-                  // Call the analytic solution to compute the temperature
-                  temperature = get_temperature_analytic(top_heat_content, min_temperature, background_temperature, temperature_, plate_velocity, effective_plate_age, adjusted_distance);
+                  double nondimensional_adjusted_distance = adjusted_distance / max_depth;
+
+                  if (apply_spline)
+                    {
+                      // A total number of (2 * spline_n_points + 1) points are picked,
+                      // spline_n_points points on each side and one at the center.
+                      // These points cover a range of (-1.0, 1.0) in adjusted_distance.
```
mibillen, line 493: change:  "in adjusted_distance"
to " in nondimensional_adjusted_distance"

mibillen (commented):
Just a few comments to clarify and then this can be merged. This is a very clean straightforward change to the code and the extra  "cost" to calculate the spline is minimal due to the way you implemented this with the get_temperatures_analytic function.

lhy11009:
I have addressed Magali's comments and rebased them.

doc/world_builder_declarations.schema.json
```diff
@@ -1168,7 +1180,11 @@
                                                                             "items": {
                                                                                 "type": "array",
                                                                                 "minItems": 1,
+<<<<<<< HEAD
```
MFraters, line 1183: This doesn't look correct. Can you run `ctest -R generate_declarations_and_schema` and check in the new result?

MFraters (approved):

MFraters:
also, do not forget to add a changlog entry, either here or in a follow up pull request.

doc/world_builder_declarations.schema.json
lhy11009: I fixed this. But this is from the conflict I get from rebase. I am not sure why this happens:

Auto-merging doc/world_builder_declarations.schema.json
CONFLICT (content): Merge conflict in doc/world_builder_declarations.schema.json
Auto-merging doc/world_builder_declarations_closed.md
CONFLICT (content): Merge conflict in doc/world_builder_declarations_closed.md
Auto-merging doc/world_builder_declarations_open.md
CONFLICT (content): Merge conflict in doc/world_builder_declarations_open.md

lhy11009:
Fix the conflicts and add change log.

MFraters:
Thanks!

lhy11009:
I have this issue I want to assert the subducting velocity as positive in the code

      WBAssertThrow(subducting_velocity >= 0, "The subducting velocity is less than 0. "
                    "Subducting velocity: " << subducting_velocity);

And then some errors are triggered. Then I realized the default value is -1.0 indicating we want to use the spreading velocity instead. 

Is there a workaround where we can still check if a incorrect negative value is entered versus if we want to use the spreading velocity instead? 

Some trials: I tried to initiate the subducting velocity as nan value, but there is a check for the input values. So this would trigger another error.

          prm.declare_entry("subducting velocity", Types::OneOf(Types::Double(NaN::DQNAN), Types::Array(Types::Array(Types::Double(NaN::DQNAN), 1), 1)),
                            "The velocity with which the ridge is moving through time, and how long the ridge "
                            "has been moving. First value is the velocity, second is the time. Default is [0 cm/yr, 0 yr]");

@MFraters ? Any idea.

PR #661, opened 2024-02-21
Add water content composition to subducting plates and oceanic plates

danieldouglas92:
Add a new composition to the subducting plate and oceanic plate features that returns the value of the water content within that feature based on the temperature and pressure. Temperature is queried within the feature and pressure is estimated by the lithostatic pressure of a user given reference density. The water content is calculated from parametrized phase diagrams for four different lithologies from a study by [Tian et. al., (2019)](https://doi.org/10.1029/2019GC008488). This study approximates PerPleX phase diagrams for sediments, Mid Ocean Ridge Basalt (MORB), gabbro, and peridotite. An example phase diagram is shown below for MORB.

image

A user defined initial water content provides an upper bound on the water within the feature, and taking the minimum of the upper bound with the calculated water content at a point yields a reasonable approximation for the water distribution within a subducting slab.

image

List of tasks:
- [x] Implement the equations for calculating the water content from Tian et. al., (2019).
- [x] Add the water content composition as an option in the oceanic plate feature
- [x] Add the water content composition as an option in the subducting plate feature
- [x] Create a test that shows a smooth distribution of water in the unsubducted oceanic plate and the subducted plate


danieldouglas92:
I've reopened this Pull Request because this now seems like the better approach over #690. 

I abandoned this approach because there was an issue where the water content would not monotonically decrease downdip of the trench, but would actually spike to large values at depth. I thought this issue would be resolved if instead of simply calculating the pointwise water content, we track the water from the trench to the point where we were calculating the water, thereby ensuring that it would never increase. This was computationally very expensive, and it turns out that the increase in water at depth was actually largely due to the fact that the parameterized phase diagrams used to determine the water content breaks down at large pressures. By capping the pressure before the parameterization breaks downs, the resulting water distribution is smooth, and largely decreases monotonically downdip. 

Below I show an example:

First, this is the temperature distribution for the example slab:

image

This is what the water content (Composition 0, in wt%) USED to look like, before limiting the pressure:

image

Notice the sudden jump in water content in the vertical portion of the slab. And with the limited pressure, this is what the water content (Composition 0, in wt%)  distribution looks like now:

image

This increase in water content is now gone.

danieldouglas92:
This ready for a look through whenever you find time for a review :) @MFraters

source/world_builder/features/subducting_plate_models/composition/water_content.cc
```diff
@@ -0,0 +1,231 @@
...
+        void
+        WaterContent::parse_entries(Parameters &prm)
+        {
+          min_depth = prm.get<double>("min distance slab top");
+          max_depth = prm.get<double>("max distance slab top");
+          density = prm.get<double>("density");
+          compositions = prm.get_vector<unsigned int>("compositions");
+          max_water_content = prm.get<double>("initial water content");
+          operation = string_operations_to_enum(prm.get<std::string>("operation"));
+          lithology_str = prm.get<std::string>("lithology");
```
MFraters, line 94: You have an enum, there is no need to store a string, since you are only really need the enum afterward. So I would suggest to convert it directly to an enum here and just pass that around. This way you only have to do a string comparison once at startup and not every time you call the plugin.

source/world_builder/features/oceanic_plate_models/composition/water_content.cc
```diff
@@ -0,0 +1,239 @@
...
+        WaterContent::~WaterContent()
+          = default;
+
+        void
+        WaterContent::declare_entries(Parameters &prm, const std::string & /*unused*/)
+        {
+          // Document plugin and require entries if needed.
+          // Add compositions to the required parameters.
+          prm.declare_entry("", Types::Object({"compositions"}),
+                            "WaterContent compositional model. Sets bound water content as a compositional field.");
```
MFraters, line 64: Could you extend this, explaining it is the Tian 2019 model, and add the changes you made to make it work? Also in the one below. Basically what you have written in the class description.

danieldouglas92:
@MFraters Thanks for the feedback! For the Tian parameterization I would say people would not want to vary the lithologies and the corresponding polynomial coefficients as these values are explicitly defined in the Tian et al., 2019 publication. However, I could certainly see someone wanting to compute the initial water content in a slab using something other than the Tian parameterization defined here (i.e. a Perplex lookup table). At the very least I can rename things to make it clear it is using the Tian parameterization, but I'll try to see if theres a way to make it more flexible to accommodate other methods such as thermodynamic lookup tables!

danieldouglas92:
@MFraters I cleaned up the code and fleshed out the documentation in places. I ended up rewording the composition name to make it explicit that this uses the parameterized phase diagrams unique to the Tian et al., 2019 publication. I decided not to try to make it generalized because unless we include thermodynamic lookup tables within the worldbuilder git repo, its not trivial to me how we would allow any user-created thermodynamic lookup table to be compatible with the code that I had already created.

MFraters (commented):
Thanks for the changes, it looks much better this way. I think it was a good call to make the pressure cutoff variable. (And a coverage of almost 98%, nice!)

There is on thing which I am a bit worried about, and that is all the vector copying you are doing inside the calculation. I think the following would make the code easier to read and more efficient since the if statement block and vector copying goes away:

header file:
```
std::array<std::vector<double>,4> LR_poly = 
{
 {-19.0609, 168.983, -630.032, 1281.84, -1543.14, 1111.88, -459.142, 95.4143, 1.97246},
{...},
{...},
{...}
}
```

This way you can access them in the .cpp file with LR_poly[periodite] or LR_poly[sediment].

so this:

```
          // Calculate the LR value from Tian et al., 2019
          for (unsigned int LR_coeff_index = 0; LR_coeff_index < LR_polynomial_coeffs.size(); ++LR_coeff_index)
            ln_LR_value += LR_polynomial_coeffs[LR_coeff_index] * (std::pow(1/pressure, LR_polynomial_coeffs.size() - 1 - LR_coeff_index));
```
would become:
```
          // Calculate the LR value from Tian et al., 2019
          for (unsigned int LR_coeff_index = 0; LR_coeff_index < LR_polynomial_coeffs.size(); ++LR_coeff_index)
            ln_LR_value += LR_poly[lithology_type][LR_coeff_index] * (std::pow(1/pressure, LR_poly[lithology_type].size() - 1 - LR_coeff_index));
```

bwt, what does LR, c and Td stand for. Can you write it out or document it? Or are those just the symbols they use in the paper?

danieldouglas92:
I didn't know you could index with an enum! Very cool, I made the changes.

MFraters (approved):
This looks great, thanks for the quick changes! I think this is ready to merge, but lets see what the testers say :)

I was doubting over one thing, and I want to document my doubt. I was not sure about whether `tian water content` be `Tian water content` or not, given that it is a name. I looked through the codebase and the only other place I could find was the gaussian temperature model (https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/main/source/world_builder/features/plume_models/temperature/gaussian.cc), where we also use a lower case. So I guess it is what we go with now :)

> @MFraters I didn't know you could index with an enum! Very cool, I made the changes.

enums in c and c++ are just integers and can be used as such (except maybe class enum). Here is a quick example: https://onecompiler.com/cpp/42prjdbr4 .

PR #664, opened 2024-02-22
Add test with a different random number seed and add comments

Wang-yijun:
The previous test uses random number seed = 1000 and this test uses random number seed = 1.


MFraters (approved):
Looks good, thanks for adding a test for this and improving the documentation. Your commit history looks really strange though. Can you squash the first three commits into one (you can leave the last one)

MFraters:
Thanks!

PR #666, opened 2024-02-22
Rename anullus to annulus

### LINKED ISSUE #539: Rename `anullus` to `annulus`?
I noticed that the gwb-grid app has an input option for the grid type called `anullus` (while the correct spelling would be `annulus`). It would be good to change it, but then I guess that would break some old files? @MFraters how would you want to handle that?

MFraters:
resolves #539


jdannberg (commented):
It's also in sphinx/user_manual/how_to_use_the_applications/gwb-grid_app.md. That's where I found it.

MFraters:
ah, thanks! I did a search and replace all files in vs code, but I still had a filter for *.cc and *.h files, so that is why it didn't show up. I fixed it now.

PR #669, opened 2024-02-23
Add tester for gwb-grid sphere.

MFraters:
follow up to #665 and part of #526 to improve coverage.


MFraters:
Note, to make this work for all testers, I had to set all depth values which where close to zero (abs < 1e-8) to zero.

PR #672, opened 2024-02-23
Add tester for gwb-grid annulus.

MFraters:
follow up to https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/665 and #669 and part of https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/526 to improve coverage.


MFraters:
It seems like this requires the same kind of fixes #669.

PR #677, opened 2024-02-26
Tracing and fixing NaN in distance_point_from_curved_planes

MFraters:
This fixes a NaN when the requested point is exactly on one of the provided trench coordinates. While tracing and fixing a issue in the utilties `distance_point_from_curved_planes` function, I created some extra asserts. I think it is still useful to leave them in the code.


MFraters:
The testcase which failed (a basic starter tutorial input and grid file) will be added in #676

Issue #670, opened 2024-02-23
Release 0.6


MFraters:
Due to wanting to submit the JOSS paper for the world builder, I would like to release version 0.6 ~~either over the weekend or early next week~~ probably end of the week, but could be later. This allow the version number to be upgraded to 1.0-pre. The 1.0 release would then coincide with the the publication of the paper.  

Here is the todo list:
- [x] run spell checker (#678)
- [x] run clang-analyzer scan-build (https://clang-analyzer.llvm.org/command-line.html) -- No bugs found.
- [x] run clang tidy (#679)
- [x] check and fix changelog entries (#681)
- [x] Update copyright year sphinx, include and header files (#684)
- [x] confirm status of cmake issues with aspect integration (#686 and https://github.com/geodynamics/aspect/pull/5592)
- [x] Confirm aspect pull request review (https://github.com/geodynamics/aspect/pull/5584)
- [x] make new zenodo doi (before the release) to include in the new release -- Done: 10.5281/zenodo.10794862
- [x] increase version number (#695)
- [x] make release
- [x] post-release version bump (#698) --> Note: Benchmarks won't work until this pull request is merged.

Please let me know if there are any other items which should be checked before the release, or which may delay it.

MFraters:
GWB v0.6.0 release complete! :tada:

PR #678, opened 2024-02-27
Add typos spell checking

MFraters:
This adds a github action for checking for typos with typos. It then suggests changes for the fixes. 

It is warned that it might go wrong if it finds typos in sections which are not changed. It should find several (it does that locally), so I am curious what will exactly happen. In general that should not be a problem though, because new typos will always be in affected lines.

related to #670


MFraters:
This functions and I have fixed all the typos. It does seem to generate more typo warning locally and in the github actions, even though they use the same version and config file. Anyway, this is a significant improvement I think of having no spell checking.

PR #679, opened 2024-02-28
Clang tidy fixes v0.6

MFraters:
I changed to the dealii .clang-tidy, given that seems more stable. The proposed changes itself are not really shocking I think.

related to #670

### DIFF
--- a/source/world_builder/kd_tree.cc
+++ b/source/world_builder/kd_tree.cc
@@ -89,8 +89,8 @@ namespace WorldBuilder
             }
 
           // Compare node's point to current closest point
-          const double distance = sqrt((node[0]-check_point[0])*(node[0]-check_point[0])
-                                       +(node[1]-check_point[1])*(node[1]-check_point[1]));
+          const double distance = sqrt((node[false]-check_point[0])*(node[false]-check_point[0])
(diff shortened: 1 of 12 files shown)


source/world_builder/kd_tree.cc
```diff
@@ -89,8 +89,8 @@ namespace WorldBuilder
             }
 
           // Compare node's point to current closest point
-          const double distance = sqrt((node[0]-check_point[0])*(node[0]-check_point[0])
-                                       +(node[1]-check_point[1])*(node[1]-check_point[1]));
+          const double distance = sqrt((node[false]-check_point[0])*(node[false]-check_point[0])
```
tjhei, line 92: Are you sure this is sensible? Isn't ``node`` an std::vector?

MFraters: I had to look this up as well, but it is an overloaded ` const double &operator[](const bool y_axis) const`:
https://github.com/GeodynamicWorldBuilder/WorldBuilder/blob/91c6b19e82a8fd6221c9ee8dd8092e765da4e345/include/world_builder/kd_tree.h#L77-L83

Maybe we should chose a better operator, or just make it a function, but clang-tidy is right in this case.

PR #680, opened 2024-02-28
Remove unused header files.

MFraters:
I tried to do this both through clang-tidy and include-what-you-use, but both also just include a whole lot of other include files and seem to clutter more than they cleanup. Here I used clangd and just went from file to file to find the unneeded includes.

related to #670


tjhei (approved):
Great, this is could be a big win.

PR #683, opened 2024-02-28
Explicitly define spherical coordinate convention

danieldouglas92:
I wasn't able to immediately find this in the place where I thought it should be. Should this maybe be added in other places in the manual where spherical geometries are being handled?


MFraters (approved):
Thanks! I agree that it wouldn't hurt if it is added to multiple places where it it relevant.

PR #686, opened 2024-02-28
CMake changes for ASPECT DebugRelease mode

tjhei:
This PR adds a number of CMake options to rename the name of the gwb library and to disable certain targets/features. This change allows us to configure a debug and a release mode build of GWB when ASPECT is configured as DebugRelease mode.

This PR should not change anything in the default configuration.


CMakeLists.txt
```diff
@@ -528,7 +551,7 @@ if(MAKE_PYTHON_WRAPPER)
     endif()
   endif()
   
-  set_target_properties(${TARGET} PROPERTIES PREFIX "_" OUTPUT_NAME "gwb")
+  set_target_properties(${WB_TARGET} PROPERTIES PREFIX "_" OUTPUT_NAME "gwb")
```
MFraters, line 554: This change seem sot be tripping it up. Note that `${TARGET}` is not equal to `WorldBuilder` but to either `_gwb` or `gwb`. 

The main problem is that the c/cpp/fortran examples expect a library called libWorldBuilder.a, which with this change is no longer generated, but it instead generated `_gwb.a`.  Commenting out the line seems to let the testers pass on my system. I think this line may have been faulty an not executed.

tjhei:
okay, seems to work now. Thank you, @MFraters . The script is somewhat messy, but I don't know of a better way to make this work within ASPECT. What do you think?
The only alternative I can think of is that you also support a ``DebugRelease`` mode (similar to deal.II where two libary files are produced). This is not a very typical thing to do and maybe not worth the headache.

CMakeLists.txt
```diff
@@ -40,6 +40,12 @@ ENDIF()
 
 set (WORLD_BUILDER_SOURCE_DIR ${PROJECT_SOURCE_DIR})
 
+# define some config options:
+set(WB_ENABLE_APPS ON CACHE BOOL "If enabled, applications gwb-grid and gwb-dat are built.")
+set(WB_ENABLE_HELPER_TARGETS ON CACHE BOOL "If enabled, targets to indenting and switching to debug/release mode are configured.")
+set(WB_ENABLE_TESTS ON CACHE BOOL "If enabled, tests are configured.")
+set(WB_ENABLE_DOCUMENTATION ON CACHE BOOL "If enabled, building of the documentation is configured.")
```
MFraters, line 47: There is already an option called `WB_BUILD_DOCUMENTATION` in `doc/CMakeList.txt`. I am fine with removing that in favor of `WB_ENABLE_DOCUMENTATION`, but that would need to be done.

CMakeLists.txt
```diff
@@ -40,6 +40,12 @@ ENDIF()
 
 set (WORLD_BUILDER_SOURCE_DIR ${PROJECT_SOURCE_DIR})
 
+# define some config options:
+set(WB_ENABLE_APPS ON CACHE BOOL "If enabled, applications gwb-grid and gwb-dat are built.")
```
MFraters, line 44: The apps are used in the tester, so if this is off and the testers are on, it needs to only run the tests which do not require either gwb-grid or gwb-dat. Can you add a github tester which turns off the apps, but leaves the tester on?

MFraters (approved):
Thanks for your work on this @tjhei! Although supporting DebugRelease in the future would be great, I don't think it needs to be done now. 

I have two small comments on the new cmake variables, but otherwise it looks good to me.

CMakeLists.txt
tjhei: the option I introduced also doesn't do anything. :-) removed.

CMakeLists.txt
MFraters: For the test, please put it in test.yml and set `needs: [test_indentation,coverage]`. If you want me to add the tester, let me know, I am happy to do it.

tjhei:
Ok, this should be good to go. Would you mind adding the additional CI step, @MFraters ?

MFraters (approved):
Sure, I will make a follow up pull request for that. Thanks for working on this!

PR #687, opened 2024-02-29
fix variable name

lhy11009:
I think this is cos_lat instead of cos_long

Also checked all the other places where longitude and latitude are mentioned and they seem all correct.


MFraters (approved):
Makes sense, thanks for the fix!

PR #690, opened 2024-03-01
[WIP] Add water content as a composition to subducting plates

danieldouglas92:
Due to a change in the implementation, PR #661 has been closed and has been repurposed in this PR. The goal is the same, but the issue with the implementation in the previous PR is that the water content at a given point within the slab did not depend on what the water content was at points updip. This new implementation accounts for the history of a given point within the slab by iterating along a plane at a fixed depth from the slab surface from the trench to the current point, comparing the water content with each iteration to the previous iteration. The method currently used is poorly optimized, but gives the desired result. 

- [x] Implement Tian et al., 2019 parametrizations
- [x] Implement water content into linear slab segments
- [ ] Implement water content into variable dip slab segments
- [ ] Create tests

### DIFF
--- a/doc/world_builder_declarations.tex
+++ b/doc/world_builder_declarations.tex
@@ -1,0 +1,7163 @@
+\section{(0) /}
+\begin{itemize}[leftmargin=0em]\item {\bf type}: object
+\item {\bf description}: Root object
+\item {\bf additionalProperties}: false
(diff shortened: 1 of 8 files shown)


danieldouglas92:
This functionality is now being merged in PR #661

PR #691, opened 2024-03-02
add missing test dependencies to run tests

tjhei:
When running all tests with ``ctest -j 10``, I often get some of the ``run_*`` tests to fail (something like "permission denied"). I think this is because we don't enforce compiling the tests before running them. This adds the missing dependencies.


MFraters (approved):
Great, thanks!

PR #704, opened 2024-04-27
Fix crash if the minimum and maximum depth are the same.

MFraters:
use the top temperature in that case.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,7 +92,7 @@ jobs:
             echo "3"
             ls /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/libexec
             echo "4"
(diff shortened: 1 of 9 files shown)


MFraters:
the coveralls stated a rate limiting, so I am just rerunning it and that seems to work for now. I will have to take a closer look at the warning though at some point. I am not sure why the doctester is failing though. I tried to fix it, but I didn't manage yet...

gassmoeller:
I will let you judge which of these testers should be fixed before merging. I think the code change itself is correct.

MFraters:
Since the readthedocs problem is unrelated to this pull request (I probably should have split the other fixes into separate pull requests to start with) and I agree that the code chagnes are correct, I will merge this and make a separate issue for the readthedocs compilation problem

PR #711, opened 2024-05-09
Update and fix most ubuntu testers to 24.04.

MFraters:
Ubuntu-latest hasn't been set yet to ubuntu-24.04, but it is available. If it works, I think it makes sense to switch the default of most tester to this. I kept a specific 22.04 tester in both debug and release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,7 @@ jobs:
     strategy: 
       fail-fast: false
       matrix: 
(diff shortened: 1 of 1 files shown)


MFraters:
It looks like the actual runners are not available yet (or at least not yet with enough quantity to start). So this will have to wait until they become available. Based on 22.04 runner release announcements it is likely to become available between now and August.

PR #712, opened 2024-05-09
Add macos 14 and only build macos 11 in mpi release

MFraters:
Macos 14 is the latest release, and also a arm version. So it would be good to see how well the tester works for this.


MFraters:
16 out of the 132 tests have issues on mac 14 arm. It seems to me mostly either small difference in the output numbers, or some small difference in the computed position caused by round off error causing it to fall on the other side of a line. It may be possible to make some the tests less sensitive to these round off errors, but I am not sure it is the case for all of them. Another option is to disable these specific tests on arm testers. 

@gassmoeller @tjhei opinions?

gassmoeller:
Disabling tests for ARM seems a bit problematic, what if they are actual bugs that only surface on arm architecture? Making the tests less sensitive to roundoff or understanding why they fail seems like the safer path (and may also have benefits for future x86 tests on new operating systems). It is more work though, and maybe not the highest priority?

tjhei:
Do you use numdiff for comparing test output right now like we do for ASPECT?

tjhei:
I would also separate Linux and osx testing.

MFraters:
Thanks for the feedback. Yes, numdiff is being used. 

I will take the longer route than in this case, although it may take a bit before I have time to address this issue.

PR #715, opened 2024-05-30
Simplify cmake C++ version setting

MFraters:
Follow up to #714, based on a comments by @tjhei and @bangerth in https://github.com/geodynamics/aspect/pull/5669.

The C++ version is already set globally, so these local c++14 marks should not have an effect. Also added a cmake setting to turn off any compiler extensions, since this reduces the portability of the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ project(WorldBuilder C CXX)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set (CMAKE_CXX_EXTENSIONS OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ project(WorldBuilder C CXX)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set (CMAKE_CXX_EXTENSIONS OFF)
```
tjhei, line 7:
```suggestion
set(CMAKE_CXX_EXTENSIONS OFF)
```

tjhei (approved):

tjhei (approved):

PR #735, opened 2024-06-01
Update documentation_style_guide.md

ljhwang:
minor fixes

### DIFF
--- a/doc/sphinx/developer_manual/contributing_to_the_documentation/documentation_style_guide.md
+++ b/doc/sphinx/developer_manual/contributing_to_the_documentation/documentation_style_guide.md
@@ -2,7 +2,7 @@
 Documentation style guide
 =========================
 
-Not only are there many ways to style the documents, there are also many ways with sphinx to reach the same output. To attempt to unify the style of the documentation and the documentation code, this page contains some general styling recommendations.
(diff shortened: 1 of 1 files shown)


doc/sphinx/developer_manual/contributing_to_the_documentation/documentation_style_guide.md
```diff
@@ -32,11 +32,11 @@ My Note
 
 # Including World Builder files
 
-When including a world builder file, it is important to be able to focus on the parts which are important, while also enabling the reader to see the  surrounding context and the whole file. Furthermore, the user should have the capability to run the models themselves and get the same output as shown. To standardize this the following structure is strongly encouraged. 
+When including a World bbilder file, it is important to be able to focus on the parts which are important, while also enabling the reader to see the surrounding context and the whole file. Furthermore, the user should have the capability to run the models themselves and get the same output as shown. To standardize this the following structure is strongly encouraged.
```
MFraters, line 35:
```suggestion
When including a World Builder file, it is important to be able to focus on the parts which are important, while also enabling the reader to see the surrounding context and the whole file. Furthermore, the user should have the capability to run the models themselves and get the same output as shown. To standardize this the following structure is strongly encouraged. 
```

PR #736, opened 2024-06-04
Remove pic flag from cmake file.

MFraters:
The -fPIC flag should be automatically set by cmake, so there is no need to specify it manually. @bangerth @tjhei

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,7 @@ if(NOT DEFINED WB_TARGET)
   set(WB_TARGET "WorldBuilder" CACHE STRING "")
 endif()
 add_library(${WB_TARGET} ${SOURCES})
+set_property(TARGET ${WB_TARGET} PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 IF(NOT CMAKE_VERSION VERSION_LESS 3.16)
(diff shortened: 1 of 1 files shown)


bangerth (commented):
That seems like the right choice.

bangerth:
I believe that the problem is that the `CMakeLists.txt` file builds a *static* library that is then later used to link into a shared library or PIC executable. There are multiple ways to achieve what you want:

* Make sure cmake knows about this, by setting the appropriate flag -- see https://stackoverflow.com/questions/38296756/what-is-the-idiomatic-way-in-cmake-to-add-the-fpic-compiler-option
* Build a shared library right away.

My personal opinion is that you should be building a shared library as it minimizes link times for downstream projects. But that's just my opinion.

MFraters:
Thanks for the advice on how to fix this. 

> My personal opinion is that you should be building a shared library as it minimizes link times for downstream projects. But that's just my opinion.

I personally like the fact that the the library is compiled into the executable, since it make the executable self contained and even allows the executable to be moved, but I can see that shorter linking times may also be nice and maybe be more useful in practice. I would suggest just fixing this now, and move the static vs dynamic library discussion into a separate issue and potentially a separate pull request.

bangerth:
Yes, these are independent questions.

bangerth (commented):
This looks correct to me now, and it also seems to work.

tjhei (approved):

Issue #738, opened 2024-06-05
combining clang and gfortran


tjhei:
I have clang, clang++, g++, and gfortran installed (not llvm flang).
When configuring as ``CC=clang-18 CXX=clang++-18 cmake -G Ninja ..``

I get
```
f95: error: unrecognized command-line option ‘-fcolor-diagnostics’
```

MFraters:
This has been fixed in #739

PR #739, opened 2024-06-05
Add linux clang tester.

MFraters:
Apparently there wasn't a linux clang tester yet, clang was only tested through mac's. This pull request adds one. Related to #738.


.github/workflows/test.yml
```diff
@@ -455,6 +455,65 @@ jobs:
       run: ctest -VV -C ${{ matrix.build_type }}
 
 
+
+  linux-clang_mpi:
+    needs: [test_indentation]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-20.04']
```
tjhei, line 464: Consider using 22.04. A newer clang will give you more warnings.

.github/workflows/test.yml
```diff
@@ -455,6 +455,65 @@ jobs:
...
+        else
+            echo "$RUNNER_OS not supported"
+            exit 1
+         fi
+      shell: bash
+
+    - name: CMake version
+      run: cmake --version
+
+    - name: GCC version
```
tjhei, line 489: Maybe show clang?

MFraters:
@tjhei I think I solved all the warnings. Can you check with your clang aspect build whether there are still warnings left?

CMakeLists.txt
```diff
@@ -371,16 +371,16 @@ if (NOT MSVC AND NOT APPLE)
...
       if(NOT ${CMAKE_VERSION} VERSION_LESS "3.9.0")
         SET(WB_COMPILER_OPTIONS_PRIVATE -pedantic -Wall -Wextra  
-        $<$<COMPILE_LANGUAGE:CXX>:-Wmost -Wconversion -Wunreachable-code -Wuninitialized -Wmissing-braces -Wunused-parameter -Wold-style-cast -Wshadow -Wfloat-equal -Wpointer-arith -Wwrite-strings
+        $<$<COMPILE_LANGUAGE:CXX>:-Wmost  -Wunreachable-code -Wuninitialized -Wmissing-braces -Wunused-parameter -Wold-style-cast -Wshadow -Wfloat-equal -Wpointer-arith -Wwrite-strings
         -Wsynth -Wsign-compare -Woverloaded-virtual -Wliteral-range -Wparentheses -Wunused-local-typedefs -Wcast-qual -fstrict-aliasing -Werror=uninitialized -Wundef 
         -Wcast-align -Wmissing-declarations -Wredundant-decls -Wdiv-by-zero -Wdisabled-optimization -Wswitch-default -Wunused>)
 
         if (${FORCE_COLORED_OUTPUT})
-          SET(WB_COMPILER_OPTIONS_PRIVATE -fcolor-diagnostics ${WB_COMPILER_OPTIONS_PRIVATE})
+          SET(WB_COMPILER_OPTIONS_PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-fcolor-diagnostics> ${WB_COMPILER_OPTIONS_PRIVATE})
```
tjhei, line 379: nice!

include/world_builder/bounding_box.h
```diff
@@ -93,7 +93,7 @@ namespace WorldBuilder
    *
    * @note The majority of this class is copied from the deal.II library.
    */
-  template <int spacedim>
+  template <unsigned int spacedim>
```
tjhei, line 96: is this unrelated or necessary? Note that we use ``int`` in ASPECT and deal.II.

MFraters: It is the easiest way to get rid of the new warnings and logically the most consistent. I don't think it will impact ASPECT in any way, since the types are not shared.

source/world_builder/kd_tree.cc
```diff
@@ -36,7 +36,7 @@ namespace WorldBuilder
                         const bool y_axis)
     {
       const size_t mid=(left+right)>>1;
-      std::nth_element(nodes.begin()+left,nodes.begin()+mid,nodes.begin()+right+1,
+      std::nth_element(nodes.begin()+static_cast<long int>(left),nodes.begin()+static_cast<long int>(mid),nodes.begin()+static_cast<long int>(right)+1,
```
tjhei, line 39: What is long int? Is this the same as long long?

source/world_builder/parameters.cc
```diff
@@ -379,7 +379,7 @@ namespace WorldBuilder
                       << base + "/" + name + "/default value");
       }
 
-    return value->GetUint();
+    return value->GetInt();
```
tjhei, line 382: That looks like an actual bug fixed.

source/world_builder/utilities.cc
```diff
@@ -1220,14 +1218,16 @@ namespace WorldBuilder
 
               data_string = datastream.str();
               filesize = data_string.size();
+              WBAssertThrow(filesize < std::numeric_limits<uint64_t>::max(),
```
tjhei, line 1221: This should be
```suggestion
              WBAssertThrow(filesize < std::numeric_limits<int>::max(),
```
Or not?

tjhei (approved):
I am not sure I understand why dim needs to be unsigned, but that's ok. I might have found one bug though.

source/world_builder/parameters.cc
MFraters: yes :)

source/world_builder/kd_tree.cc
MFraters: no, a `long int` is the same as a `long` and a `long long int` is the same as a `long long` (I also had to look it up here: https://en.cppreference.com/w/cpp/language/types). This is apparently what nodes.begin() returns (it is a difference type).

source/world_builder/utilities.cc
MFraters: hmm, filesize above is a uint64_t, but in the block below it is a filesize is an int (probably unsigned because string.size() returns an size_t). I don't think it matters much, but I will make it all an int in this block.

tjhei: well, filesize is a 64 bit value, so it is alway less than the maximum value you can store...

source/world_builder/utilities.cc
```diff
@@ -1219,7 +1217,9 @@ namespace WorldBuilder
                 }
 
               data_string = datastream.str();
-              filesize = data_string.size();
+              filesize = static_cast<int>(data_string.size());
```
tjhei, line 1220: this is incorrect. That way the Assert never triggers.

MFraters:
@Timo Thanks for catching that, I have addressed your comment :)

PR #740, opened 2024-06-05
fix disabling fortran

tjhei:
I have fortran, but I am disabling it using ``WB_MAKE_FORTRAN_WRAPPER=OFF``.

Without this patch I am getting
```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_Fortran_COMPILE_OBJECT
```

also see https://github.com/geodynamics/aspect/pull/5792/files


MFraters (approved):
Looks good to me, thanks!

Issue #528, opened 2024-01-24
Add velocity plugin system


MFraters:
For numerical models it can be useful to have for example velocity boundary conditions defined per tectonic feature. This would require 
1. to add an extra output interface, similar to temperature, compositions or grains. The default output should probably be zero. 
2. Add a new plugin system in the .cc files in `source/world_builder/features/` (except for interface.cc)
3. Add at least one plugin to `source/world_builder/features/*/velocity/`

The part which needs to be thought out is how to design the input. In Cartesian coordinates, providing a single vector and/or providing a function should work well enough. In spherical coordinates, I think the most useful representation is probably an euler pole. I don't directly see a use case for providing a single vector in spherical coordinates, but I am happy to be convinced. If this would be the case, then the plugin would interpretate the single vector in Cartesian coordinates as a Cartesian vector to apply in that feature, and in spherical coordinates as  an euler pole, and compute the Cartesian velocity vector at the requested location.

It could look something like this:
```json
{
    "velocity models":
     [
         {"model":"uniform", "vector":[1,2,3], "max depth":100e3},
         {"model":"uniform", "vector":[-1,-2,-3], "min depth":100e3}
     ]
}
```

Another possible plugin would be to provide an ascii-data plugin, which could be used to for example prescribe the result of a global model to a regional model.

MFraters:
An other possible model would be a vertical compensating model, where a velocity is specified between certain depth ranges and the resulting "inflow" or "outflow" would be compensated by an equivalent in- or outflow, with a transition zone between the two if desired. 

It would allow for easily set and modify the following boundary condition: 
![image](https://github.com/user-attachments/assets/8c83ce11-b53a-405f-aa53-021fc98041f7)

It could look something like this:
```
{"model":"vertical compensate", "min depth": 0, "max depth":500e3, "max transition length":10e3,
  "forcings":[{"vector":[1,2,3],"min depth":0, "max depth":100e3}]}
```

MFraters:
I have a proof of concept implementation. It is still very rudimentary, but it can output velocity (both gwb-grid and gwb-dat) using a velocity modules for each feature.

![image](https://github.com/user-attachments/assets/3b00dca4-b758-4a80-b9c0-6f6954143c27)

PR #761, opened 2024-10-13
[WIP] add velocity infrastructure.

MFraters:
This pull request implements #528. It is still very rudementary and a work in progress. 

I am thinking of renaming the current `uniform` models to `raw`, since they just work with raw velocity input. I think the `uniform` model should be more useful, given that it is usually the simple default model. I think uniform in this case should not make the same value everywhere (which is would raw (constant?) would be, but for example uniform in continent would provide a direction, dip and magnitude (which is more like how the rest of the models work, 2d top down + dip) and for a slab or fault it would be relative to the slab geometry. In spherical uniform should also be relative to the geometry.

Example of current functionality: 
image


MFraters:
This is almost ready to merge I think. It also contains some other items, including an option to replace the reference results from any test that fails, and limiting the resolution of the doc_* tests to improve tester performance.

PR #767, opened 2024-11-12
Convert water content from percent to a fraction in tian2019 composition model

danieldouglas92:
I ran into this issue a while ago and I fixed it locally and forgot to open a PR, but I ran into it again just now so I am making sure this is addressed. Within the WorldBuilder the water content output by the tian2019 composition model is in % (i.e. 2%), but ASPECT requires a mass/volume fraction and so for this composition feature to be useful this must be converted to a fraction (i.e 0.02). This PR divides the final value by 100 so that when it is fed to geodynamic modeling codes it is a fraction and not a percent.


MFraters (commented):
I have no objections against this, but can you add a changelog?

Also, is this not covered by tests, since no tests change?

danieldouglas92:
@MFraters Yes I'll definitely add a changelog, and there is a test (`gwb-dat/water_content_subducting_plate`) that I'll fix, it's giving a 0 water content everywhere for some reason

danieldouglas92:
The test results were updated in PR #761, but I looked at what they were before this and confirmed that all the water contents are the same as what they were originally (except now being divided by 100).

MFraters (approved):

Issue #766, opened 2024-11-07
[Tracking Issue] Improve Parameter Documentation


danieldouglas92:
My lab-mate @PrajaktaPMohite and I were working on a worldbuilder file that included faults for one of her models and when we looked at the manual in the section called `Parameter Documentation/Features` we noticed that this section is almost entirely empty (except for the `Plume` feature). 

Similar to the developer / API documentation issue I thought I would create an issue for this part of the manual so that there is a reminder that this needs to be completed as well.

- [ ] Parameters / Features
    - [ ] Oceanic Plate
    - [ ] Subducting Plate
    - [ ] Mantle Layer
    - [ ] Continental Plate
    - [ ] Faults

@MFraters Is the idea that each of these features will be expanded into another drop down list showcasing the various temperature and composition models that they support as well?

MFraters:
Yes, that is a good point. Thanks for creating this tracking issue! You can also add the plume to the list and check if off for completeness :)

> @MFraters Is the idea that each of these features will be expanded into another drop down list showcasing the various temperature and composition models that they support as well?

Yes, the idea of this sections is that if you quickly want to know how a feature works and what parameteres there are, this would be easier to read then the actual full parameter documentation. Although I don't know what format would work best for this, so if you or @PrajaktaPMohite want to give it a shot, feel free to experiment!

PR #768, opened 2024-11-19
Add Parameter Documentation for Subducting Plate Feature

danieldouglas92:
Update the manual by adding the section for the subducting plate feature in parameter documentation, addresses a point in Issue #766.


MFraters (commented):
Thanks for writing this up, very nice! Also cool to see how you made the figures to explain the concepts. I have a few small comments, but otherwise looks good to me.

1. I guess you are trying to get into math mode with with the $ around some variables, but it doesn't render correctly. Maybe we to enable an extra plugin in sphinx?
2. Currently there is a limitation that the number of segements need to be the same in all sections. Could you add that?
3. In the downdip_sections.png figure, the black cross-sections are a bit hard to see. Maybe a lighter color or an outline would help.

danieldouglas92:
@MFraters Thanks for the quick review! I addressed points 1 and 2, I just made M and N italicized, I don't think we need to worry about the math mode extension for sphinx at least for this PR. For point 3 I'll remake the figures later tonight, I also think the font should be a little larger, so I will comment at you again when it's ready for you take another look.

MFraters:
@danieldouglas92 Is this ready for another review?

danieldouglas92:
@MFraters Yes it is!

MFraters (approved):
Thanks for writing this up!

Issue #772, opened 2024-12-10
MPI issue on Ubuntu 24.04


TNeumann98:
Prerequisite for Ubuntu version 24.04.1 (LTS x86-64, Ubuntu-noble), one needs to pre-install libopenmpi-dev via
$sudo apt install libopenmpi-dev

MFraters:
resolved by #776

PR #776, opened 2024-12-11
Update installation instructions for Linux

alanjyu:
Add the required modules for compiling the WorldBuilder. Address Issue [#772](https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/772).


MFraters (changes requested):
Thanks for making this pull request! Could you move point 3 to the first position? It is always good to begin with updating you system before doing anything else (especially on WSL it seems). 

I am still not sure about why libopenmpi-dev is needed, and that most likely has more to do with the detection through cmake. But that would require a bit more investigation, so for now I think it is fine to just recommend to install it.

alanjyu:
Done!

MFraters (approved):
Thanks, looks good now!

PR #788, opened 2025-01-20
Fix 2D index for data

Djneu:
This is a very small change that updates the data output indexes for compositions in 2D.


MFraters (approved):
Thanks! Looks good.

PR #783, opened 2025-01-20
Added my developer name in the author list

alanjyu:
I made a change to the Author's list. This is for a Git tutorial during the GWB Hackathon 2025.

Issue #778, opened 2025-01-16
continental temperature from Chapman (1986)


MFraters:
In the ASPECT coninental extension cookbook (https://github.com/geodynamics/aspect/blob/main/cookbooks/continental_extension/continental_extension.prm), there is a continental temperature which I think would be interesting to add to the geodynamic world builder. 

Taken from the text from the cookbook:
```
# Typical continental geotherm based on equations 4-6 from:
#   D.S. Chapman (1986), "Thermal gradients in the continental crust",
#   Geological Society of London Special Publications, v.24, p.63-70.
# The initial constraints are:
#   Surface Temperature  - upper crust (ts1) = 273 K
#   Surface Heat Flow    - upper crust (qs1) = 0.055 W/m^2
#   Heat Production      - upper crust (A1)  = 1.00e-6 W/m^3;
#   Heat Production      - lower crust (A2)  = 0.25e-6 W/m^3;
#   Heat Production      - mantle (A3)       = 0.00e-6 W/m^3;
#   Thermal Conductivity - all layers        = 2.5 (W/(m K));
# To satisfy these constraints, the following values are required:
#   Surface Temperature  - lower crust (ts2) = 633 K
#                        - mantle (ts3)      = 893 K
#   Surface Heat Flow    - lower crust (qs2) = 0.035 W/m^2;
#                        - mantle      (qs3) = 0.030 W/m^2;
# Note: The continental geotherm initial temperature model
#       plugin can be used to compute an identical geotherm
#       for the lithosphere. An example of how to use this
#       plugin is illustrated in the test for this cookbook
#       (tests/continental_extension.prm).
```

PR #790, opened 2025-01-20
Add elevation output for each Cartesian point

alanjyu:
Added an output case (case 7) in world.cc, then for the variable calculate the elevation (minimum depth) value at each Cartesian point.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@ There have been contributions on the development from:
   Chris Mills,
   Arushi Saxena,
   Wim Spakman,
-  Cedric Thieulot 
+  Cedric Thieulot,
+  Yijun Wang,
   and
-  Yijun Wang
+  Alan Yu
--- a/source/world_builder/features/continental_plate.cc
+++ b/source/world_builder/features/continental_plate.cc
@@ -267,7 +267,7 @@ namespace WorldBuilder
                         grains.unroll_into(output,entry_in_output[i_property]);
                         break;
                       }
-                      case 4:
+                      case 4: // tag
                       {
                         output[entry_in_output[i_property]] = static_cast<double>(tag_index);
                         break;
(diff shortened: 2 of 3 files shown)


source/world_builder/features/continental_plate.cc
```diff
@@ -295,6 +295,20 @@ namespace WorldBuilder
                         //std::cout << "vel=" << output[entry_in_output[i_property]] << ":" << output[entry_in_output[i_property]+1] << ":" << output[entry_in_output[i_property]+2] << std::endl;
                         break;
                       }
+                      case 7: // elevation
+                      {
+                        if(!std::isnan(output[entry_in_output[i_property]])){
```
gassmoeller, line 300: Are you sure this should not be:
```suggestion
                        if(std::isnan(output[entry_in_output[i_property]])){
```

Do you want to check here if the output is not a number in order to then set it to some number?

AUTHORS
```diff
@@ -13,6 +13,7 @@ There have been contributions on the development from:
   Chris Mills,
   Arushi Saxena,
   Wim Spakman,
-  Cedric Thieulot 
+  Cedric Thieulot,
+  Yijun Wang,
   and
-  Yijun Wang
+  Alan Yu
```
gassmoeller, line 19: I think you created your branch from your other branch (the one used for #783) that is why this change shows up in this PR. Find me to talk about how to fix this.

source/world_builder/features/continental_plate.cc
```diff
@@ -295,6 +295,20 @@ namespace WorldBuilder
                         //std::cout << "vel=" << output[entry_in_output[i_property]] << ":" << output[entry_in_output[i_property]+1] << ":" << output[entry_in_output[i_property]+2] << std::endl;
                         break;
                       }
+                      case 7: // elevation
```
gassmoeller, line 298: this may be unrelated to your pull request, but why is elevation 7 and not 6? Why is 6 missing?

gassmoeller (commented):
The change itself looks ok to me (to add elevation output), but I have a few questions before proceeding with this PR, see my comments.

source/world_builder/features/continental_plate.cc
alanjyu: Case 6 is handled by @Djneu and is related to the density of the compositional field.

source/world_builder/features/continental_plate.cc
alanjyu: That was a mistake. This is addressed in Commit [0599baa](https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/790/commits/0599baaa1902a24c10aec7df2cff1d15bc7b4252). Thank you!

AUTHORS
alanjyu: I have closed that pull request (used for a tutorial).

MFraters (commented):
Given that #778 has been merged, I think I would be best to reopen #783 because that can get merged easily and updating the authors list is not really related to  this pull request (Add elevation output for each Cartesian point). I do feel there is some stuff missing here which I thought you worked on. Did you not push that, or do I misremember?

alanjyu:
I have implemented and tested this feature. However, an issue came up where if you set the feature min depth to something negative, the composition min depth is still set to zero. During the test, a point at negative depth (positive height) will not pick up the composition field unless its min depth matches that of the feature. I am still working on that fix, then will push this along with the test model.

MFraters:
Thanks for the update. Let me know if you want to have a chat about the issue on Monday or Tuesday

PR #792, opened 2025-01-20
Disable automatic MPI linkage

gassmoeller:
This change disables automatically looking for MPI and link to it if found. Instead now the user needs to manually set the USE_MPI cmake variable in order to make cmake look for the MPI package.

The rationale for this change is the bug (or unintended interaction) described in https://gitlab.kitware.com/cmake/cmake/-/issues/18196. If on a debian system the apt package `openmpi-bin` is installed, but `libopenmpi-dev` is not installed, then the cmake call to `find_package(MPI)` crashes with an error message that is incomprehensible to new users. We cannot test for this and write a helpful error message. And since we dont need MPI for most cases the safest option is to disable linking to MPI by default, and emphasize in the documentation that both MPI runtime and MPI headers have to be installed in order to use MPI.

Edit: I should probably also add documentation on how to enable the USE_MPI variable in the installation process. I will add this tomorrow, it is getting late.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 - The tian2019 composition model now returns a mass fraction instead of a mass percentage. \[Daniel Douglas; 2024-11-12; [#767](https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/767)\]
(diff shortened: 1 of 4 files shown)


gassmoeller:
Ok, I added installation instructions for the MPI parallelization. This is ready for a review.

MFraters (approved):
Looks good to me, thanks for fixing this!

PR #794, opened 2025-01-20
[WIP] Add density structure to world builder

Djneu:
Here is the initial try at this with an added continental plate model. So far, in the uniform model you give the number of compositions and then a list of densities for each composition.

For the calculation, I removed the max min depth, as it doesn't seem necessary when the density only depends on the composition. But maybe that will make things easier later on with multiple features?

### DIFF
--- a/include/world_builder/features/continental_plate.h
+++ b/include/world_builder/features/continental_plate.h
@@ -50,6 +50,10 @@ namespace WorldBuilder
       {
         class Interface;
       }  // namespace Temperature
+      namespace Density
+      {
+        class Interface;
+      }  // namespace Density
     }  // namespace ContinentalPlateModels
 
     /**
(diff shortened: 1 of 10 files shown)


Djneu:
I've updated the density files and added a test (hopefully in the correct location). I wasn't able to set densities to nan initially, as it would hang during runtime with the current line of code. Is there a different way to set this since densities is a vector and not a double?

MFraters (commented):
Sorry for the late review. I think this is almost ready. 

I don't think we should always output all the available properties, but make them parameters in the .dat file instead. Something like `# density = true`.

I am doubting a bit whether to include the density interface in world.h. The idea is to move to the properties interface, but having a direct interace can be really nice, even though it might not be as efficient and scalable...

There is currently a merge conflict, could you rembase?

Djneu:
Here is an image of the 5 grid tests added for this pull request, visualizing the density field.

image

PR #796, opened 2025-01-20
Add some documentation within bezier_curve.cc

danieldouglas92:
A little more documentation could probably be added around the section of code where the side of the line that each control point lies on is calculated, but I wanted to open this for now to get feedback on what I wrote.

### DIFF
--- a/source/world_builder/objects/bezier_curve.cc
+++ b/source/world_builder/objects/bezier_curve.cc
@@ -32,6 +32,10 @@ namespace WorldBuilder
 {
   namespace Objects
   {
+    /*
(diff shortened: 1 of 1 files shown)


source/world_builder/objects/bezier_curve.cc
```diff
@@ -119,7 +132,7 @@ namespace WorldBuilder
             }
           }
 
-          for (size_t p_i = 1; p_i < n_points-1; ++p_i)
+          for (size_t p_i = 0; p_i < n_points-1; ++p_i)
```
MFraters, line 135: I think this is breaking the tests. Was this change intentional? If so, since this pull request is just about documentation changes, it would best go in a different pull request.

PR #801, opened 2025-01-21
Fix indent warning

gassmoeller:
Running the indent script used to produce the following warning for a while now:

xargs: warning: options --max-args and --replace/-I/-i are mutually exclusive, ignoring previous --max-args value

This PR fixes the warning by removing the `-n` (max-args) option that was ignored anyway. It also takes over some improvements from https://github.com/geodynamics/aspect/pull/6153 to print file names separated by NULL characters to avoid problems with spaces and special characters in file names.

### DIFF
--- a/doc/indent
+++ b/doc/indent
@@ -34,9 +34,9 @@ fi
 # collect all header and source files and process them in batches of 50 files
 # with up to 10 in parallel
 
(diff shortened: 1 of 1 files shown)


MFraters (approved):
Thanks!

PR #803, opened 2025-01-21
Typo fix in: Update 16_mass_conserving.md

SabatinoSantangelo:
"developed" instead of "develop" at line 5


alarshi:
Thanks for the fix! Looks good to me

PR #807, opened 2025-01-21
Update doxygen design

gassmoeller:
This is an update to the doxygen html stylesheet similar to geodynamics/aspect#6094. The new design is more modern, but also in particular scales better to mobile devices and other resolutions, and supports a switch between light and dark theme.

Old design:

Screenshot from 2025-01-21 18-11-27

New design:

Screenshot from 2025-01-21 17-56-41


MFraters (approved):

MFraters:
I forgot to ask, can you make a changlog entry (and a hackathon entry)?

PR #812, opened 2025-01-22
Create a Cookbook of a Hydrated Slab

danieldouglas92:
Create a cookbook which generates a 2D subducting plate in a Cartesian geometry and using the `tian2019` composition model to generate the initial bound water within the subducting plate.


MFraters (approved):
This looks great, thanks!

PR #814, opened 2025-01-23
Add interface for topography models.

Djneu:
Here is adding the interface for having topography models and defining them in the .wb file. So far the model is named uniform, though I think this will be changed later on to something better fitted to the lowest feature topography model.

Next steps would be:

1. Make a function that allows world.cc to return the surface type when called from an outer program.
2. Combine uniform.cc with an actual topography model.


include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,144 @@
+/*
+  Copyright (C) 2018-2024 by the authors of the World Builder code.
```
gassmoeller, line 2: 2025

include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,144 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef WORLD_BUILDER_TOPOGRAPHY_MODELS_INTERFACE_H
+#define WORLD_BUILDER_TOPOGRAPHY_MODELS_INTERFACE_H
```
gassmoeller, line 21:
```suggestion
#ifndef WORLD_BUILDER_TOPOGRAPHY_INTERFACE_H
#define WORLD_BUILDER_TOPOGRAPHY_INTERFACE_H
```

include/world_builder/parameters.h
```diff
@@ -275,6 +280,14 @@ namespace WorldBuilder
        */
       std::unique_ptr<WorldBuilder::GravityModel::Interface> gravity_model;
 
+      /**
+       * A pointers to the gravity model. This variable is responsible for
```
MFraters, line 284: Could you search for gravity and replace it with topography?

include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,144 @@
...
+
+#include "world_builder/parameters.h"
+#include "world_builder/types/string.h"
+
+
+namespace WorldBuilder
+{
+  class World;
+
+  namespace TopographyModel
```
gassmoeller, line 33: The use of `model` attached to every plugin system is something that bothers me in ASPECT as well as here. It makes sense in some cases where the first word seems to demand for it (as in `material model`, `material` alone is just not very expressive), but in many other cases we got rid of it (as in initial temperature model -> initial temperature). I would be in favor of not introducing `model` for the topography.

```suggestion
  namespace Topography
```

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace WorldBuilder
+{
+
+  namespace TopographyModel
+  {
+    /**
+     * Register header file
+     */
+    //WB_REGISTER_COORDINATE_SYSTEM_HEADER(Uniform)
```
MFraters, line 36: Could you remove this?

include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,144 @@
...
+
+        /**
+         * declare and read in the world builder file into the parameters class
+         */
+        virtual
+        void parse_entries(Parameters &prm) = 0;
+
+        /**
+         * declare and read in the world builder file into the parameters class
+         */
```
gassmoeller, line 71: update comment

include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,144 @@
...
+      { \
+        Interface::registerType(#name, klass::declare_entries, this); \
+      } \
+      std::unique_ptr<Interface> create(World *world) override final { \
+        return std::unique_ptr<Interface>(new klass(world)); \
+      } \
+  }; \
+  static klass##Factory global_##klass##Factory;
+
+  } // namespace CoordinateSystems
```
gassmoeller, line 141: topography

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
+/*
+  Copyright (C) 2018-2024 by the authors of the World Builder code.
```
gassmoeller, line 2: all copyright notices: 2025

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef WORLD_BUILDER_TOPOGRAPHY_MODEL_UNIFORM_H
+#define WORLD_BUILDER_TOPOGRAPHY_MODEL_UNIFORM_H
```
gassmoeller, line 21:
```suggestion
#ifndef WORLD_BUILDER_TOPOGRAPHY_UNIFORM_H
#define WORLD_BUILDER_TOPOGRAPHY_UNIFORM_H
```

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+namespace WorldBuilder
+{
+
+  namespace TopographyModel
+  {
+    /**
+     * Register header file
+     */
+    //WB_REGISTER_COORDINATE_SYSTEM_HEADER(Uniform)
+
```
gassmoeller, line 37: delete

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * This implements a Uniform geometry model.The Uniform geometry model
+     * doesn't do anything with the coordinates, but is needed to have a common
+     * interface for all the geometry models.
```
gassmoeller, line 42: topography

MFraters (commented):
I am on two minds about the naming of the class. Originally I thought just naming it none, to indicate that there is no topography. But uniform, with a default of zero would also work. Opinions?

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+     */
+    class Uniform final : public Interface
+    {
+      public:
+        /**
+         * constructor
```
gassmoeller, line 48:
```suggestion
         * Constructor
```

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+        static
+        void declare_entries(Parameters &prm, const std::string &parent_name = "");
+
+        /**
+         * declare and read in the world builder file into the parameters class
+         */
+        void parse_entries(Parameters &prm) override final;
+
+        /**
+         * declare and read in the world builder file into the parameters class
```
gassmoeller, line 69: update

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+         */
+        std::string
+        get_surface_type(Parameters &prm) override final;
+
+      private:
+
+        /**
+         * Whether to consider a reference or deformed surface.
+         */
+        const std::string  surface_type = "reference_surface";
```
gassmoeller, line 79: while a string is flexible, it is also quite expensive to use. In addition it could contain arbitrary new values. I would suggest to introduce an `enum` type in the public part of this class (or even outside the class, but inside the `Topography` namespace. Something like:

```
  enum SurfaceType
  {
    reference_surface,
    deformed_surface};
```

This also guarantees that only these two values can be used.

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,85 @@
...
+
+    };
+  } // namespace CoordinateSystems
```
gassmoeller, line 82: Topography

include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * This class is an interface for the specific coordinate systems.
+     */
+    class Interface
+    {
+      public:
+        /**
+         * constructor
+         */
+        Interface();
```
gassmoeller, line 47: since the constructor is the default constructor anyway you can remove this declaration.

include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,144 @@
...
+
+        /**
+         * Destructor
+         */
+        virtual
+        ~Interface();
```
gassmoeller, line 53: If you do the following you can remove the desctructor from the interface.cc file:

```suggestion
        virtual
        ~Interface() = default;
```

source/world_builder/topography/uniform.cc
```diff
@@ -0,0 +1,67 @@
...
+      = default;
+
+    void
+    Uniform::declare_entries(Parameters &prm, const std::string & /*unused*/)
+    {
+      // Add depth method to the required parameters.
+      prm.declare_entry("", Types::Object(),
+                        "Uniform gravity model. It returns the gravity vector in a Cartesian coordinate system at "
+                        "a given position, which has a constant magitude for the whole domain. The vector points "
+                        "down in cartesian coordinates and to the center of the sphere in spherical coordinates.");
```
gassmoeller, line 45: nothing is parsed below so nothing needs to be declared here either?

gassmoeller (commented):
Yes, the structure looks correct to me, but I have a number of suggestions for naming and you will have to update a bunch of comments.

include/world_builder/topography/interface.h
Djneu: I updated this to topography class everywhere, is that the change that was supposed to be made?

source/world_builder/topography/uniform.cc
Djneu: Removed the declaration.

Djneu:
Hi both, I think I've addressed all the previous comments. Next I think I'll need to add a case 7 to world.cc and make sure calling get_topography will return 0 and that I can also return the correct surface type with the get_surface_type function.

include/world_builder/topography/interface.h
```diff
@@ -0,0 +1,157 @@
...
+
+        /**
+         * declare and read in the world builder file into the parameters class
+         */
+        virtual
+        SurfaceType
+        get_surface_type() = 0;
+
+        /**
+         * Returns a temperature based on the given position, depth in the model,
```
anne-glerum, line 95:
```suggestion
         * Returns a topography based on the given position, depth in the model,
```

source/world_builder/world.cc
```diff
@@ -532,8 +546,6 @@ namespace WorldBuilder
     return properties(point, depth, {{{2,composition_number,0}}})[0];
   }
 
-
-
   WorldBuilder::grains
```
gassmoeller, line 549: why did you delete these lines?

include/world_builder/topography/uniform.h
```diff
@@ -0,0 +1,87 @@
...
+        static
+        void declare_entries(Parameters &prm, const std::string &parent_name = "");
+
+        /**
+         * declare and read in the world builder file into the topography class
+         */
+        void parse_entries(Parameters &prm) override final;
+
+        /**
+         * declare and read in the world builder file into the topography class
```
anne-glerum, line 63: These comments are all the same, which makes them less informative. Maybe make them more specific to the functions?

gassmoeller (commented):
some comments from my previous review are not addressed yet. but generally closer to being ready.

source/world_builder/topography/uniform.cc
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Returns a topography based on the position. In the uniform model,
+     * we assume there is no topography.
+     */
+    double
+    Uniform::get_topography(const Point<3> &/*position*/,
+                            const Objects::NaturalCoordinate &/*position_in_natural_coordinates*/,
+                            const double /*depth*/) const
```
MFraters, line 59: Do you expect the depth to be needed? Do you forsee that the returned topography would be different based on the depth?

PR #815, opened 2025-01-23
Update classname in macro

gassmoeller:
@MFraters: As discussed, maybe a minor change, but I would prefer if we can use `classname` instead of `klass` as the argument for these macros. `classname` follows more closely good English language practice and includes a bit more meaning as well (and it is what we use in ASPECT for these macros). `klass` was just used before to avoid the use of `class` which is a c++ keyword.


MFraters (approved):
Thanks!

PR #817, opened 2025-01-23
Update benchark actions/upload-artifact to v4

MFraters:
Otherwise the benchmarks automatically fail due to v3 being deprecated.


gassmoeller (approved):
This is good, but we can also see if #816 will fix it using a bot PR. Let's keep this open for a bit before merging.

MFraters:
It ran depenabot which is great. Let's merge this since it is already done and we need to fix this quickly.

PR #826, opened 2025-01-23
Added content to the Fortran interface section in the API_manual part

olihen:
It would be great if someone could quickly read it, to verify if it would make sense for other users


MFraters (commented):
Thanks for writing this up! I can do a more detailed review later, but a few comments as discussed. 

1. Could you breakup the chick part into a new page, similair to c++ and aspect? You can mention that page in here so people know that it is there.
2. a hint for the `has_output_dir` and `output_dir`, if the `has_output_dir` variable is true, the world builder lib will be allowed to write files in that directory. Currently it is just writing the world builder json schema and declarations file to this directory.

PR #829, opened 2025-01-24
[WIP] Add Contours as a Method for Creating Subducting Plate Features

danieldouglas92:
Begin working towards addressing issue #429. This is still an early work in progress, and a lot of the necessary parameters are currently hard-coded in, but this is laying the ground work to allow the user to define planar features like faults and slabs by prescribing the location of depth contours.

The current implementation requires defining the depth of the contour, and the coordinates of the depth contour projected onto the surface of the world. The coordinates of each contour is fit with a bezier curve, the total arclength of the bezier curve is determined and is then divided up into N segments of even length. Finally, the segments are iterated over, and are connected piecewise linearly to the corresponding segment on each depth contour, creating a 2-D planar feature in 3-D space. A thickness is provided, which then yields a 3-D subducting plate or fault.

After this is generalized in such a way that the user can actually provide the depth contours within the world builder file, this could hypothetically be extended to allow for the contours to be connected in a higher order way (i.e. quadratic), instead of piece wise linearly, depending on how jagged the feature is with this current method.

Below I show two examples of how this works for very simple geometries. This shows a slab that is made up of 2 depth contours, and in the top picture the 2nd depth contours is shallower than the 2nd depth contour in the bottom picture. 

image

image

### DIFF
--- a/include/world_builder/utilities.h
+++ b/include/world_builder/utilities.h
@@ -487,6 +487,14 @@ namespace WorldBuilder
     std::vector<double>
     calculate_effective_trench_and_plate_ages(std::vector<double> ridge_parameters, double distance_along_plane);
 
+    /*
+     *
+     * @brief Calculate the distance of a point to a the closest point on a rectilinear plane
+     * @param p1 The first point defining the plane
+     * @param p2 The second point defining the plane
+     * @param p3 The third point defining the plane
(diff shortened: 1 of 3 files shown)


MFraters (commented):
Cool! Let me know when you want a review of the code :)

PR #830, opened 2025-01-24
Add temperature models that reads from ascii data file

jdannberg:
I added a temperature model for oceanic plates that can read from an ascii data file, and I also made a test and a cookbook. 

Not sure when you'll get to reviewing this, but I just wanted to upload the state I got to. The cookbook does not have any text or images, but other than that, everything should work. 

This is how the cookbook output looks like:
Screenshot from 2025-01-24 21-48-37
I obviously can't exactly match the edges of the plate (which uses a T-dependent conductivity) and the slab (which uses mass-conserving, and therefore a constant conductivity), but I tried to make it close.

For now I've simply added the python script into a contrib folder, but let me know if you want that somewhere else.


include/world_builder/features/oceanic_plate_models/temperature/ascii_data.h
```diff
@@ -0,0 +1,130 @@
...
+            };
+
+            TableData age_depth_table;
+
+            /**
+             * A function that reads in the ascii data file that provides
+             * the temperature distribution in dependence of age and depth
+             * and returns it as TableData.
+             */
+            bool readDataFromFile(const std::string &filename,
```
MFraters, line 111:
```suggestion
            bool read_data_from_file(const std::string &filename,
```

include/world_builder/features/oceanic_plate_models/temperature/ascii_data.h
```diff
@@ -0,0 +1,130 @@
...
+                                  TableData &data) const;
+
+            /**
+             * A function that extracts the temperature from the table structure
+             * by linearly interpolating between the closest entries for depth
+             * and age.
+             */
+            double interpolateTemperature(const TableData &table,
```
MFraters, line 119:
```suggestion
            double interpolate_temperature(const TableData &table,
```

source/world_builder/features/oceanic_plate_models/temperature/ascii_data.cc
```diff
@@ -0,0 +1,388 @@
...
+                    {
+                      spreading_rates_for_ridge.push_back(spreading_velocities.second[ridge_point_index]);
+                    }
+                  ridge_point_index += 1;
+                }
+              spreading_velocities_at_each_ridge_point.push_back(spreading_rates_for_ridge);
+            }
+
+          // Read in the data table
+          if (!readDataFromFile(data_directory + data_file_name, age_depth_table))
```
MFraters, line 136:
```suggestion
          if (!read_data_from_file(data_directory + data_file_name, age_depth_table))
```

MFraters (approved):
Thanks for contributing this! I think the contrib folder makes sense for now, give that @Minerallo is also placing his python scripts there. I think the folder name can be a bit more specific though than python. Maybe `1D-diffusion-table`, or something like that?

I have a few small suggestions to make the testers happy and to change camel case into snake case for functions. I think we should also separate out the ascii data functionality into a generic utilities function, but we can do that later.

jdannberg:
Great, thank you! I've addressed your comments.

PR #835, opened 2025-02-14
Fix Asserts and typos

anne-glerum:
I ran into some Asserts when converting a 3D .grid file into a 2D one. There were some inconstistencies in the Asserts checking the number of grid cells in each direction and some typos that this PR fixes.

### DIFF
--- a/source/gwb-grid/main.cc
+++ b/source/gwb-grid/main.cc
@@ -354,7 +354,7 @@ int main(int argc, char **argv)
       if (find_command_line_option(argv, argv+argc, "-h") || find_command_line_option(argv, argv+argc, "--help"))
         {
           std::cout << "World Builder Grid Visualization tool.\n"
-                    <<  "This program loads a world builder file and generates a visualization on a structured grid "
+                    <<  "This program loads a World Builder file and generates a visualization on a structured grid "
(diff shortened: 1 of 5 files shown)


MFraters (approved):
Thanks for finding and fixing this! I am supprised that this is not tested (I guess because the lines are technically "covered"). 

Could you fix the failing tests (update the message in the log)?

tests/gwb-grid/grid_help.log
tests/gwb-grid/grid_non_existing_wb.log

Thanks!

anne-glerum:
Sure! I fixed the tests by hand, as I couldn't find a changediff artifact to download and apply. Is there such a thing produced by the tester?

MFraters:
Thanks for fixing it. The remaining issue had to with some of the postprocessing of the results, to get rid of run specific info (for example, number of threads, run paths, etc.). I pushed a commit which fixes that. 



> Sure! I fixed the tests by hand, as I couldn't find a changediff artifact to download and apply. Is there such a thing produced by the tester?

Unfortunately not yet. It has been on my todo list to add since the hack. It might not be much work, so I will give it a go :)

MFraters:
@anne-glerum See #837

anne-glerum:
Thanks for the fix and the artifact!

PR #842, opened 2025-08-19
Add litho1.0 to the world builder

MFraters:
This add a string option to the min and max depth so that you can select to use a lith1.0 layer (lithosphere, crust 3, crust 2, etc.). 

My original approach was to let an external program make this directly using the litho1.0 provided application. This is not very convient, but it allows you to tweak the output yourself. 

My second approach was to create a seperate composition plugin where you would assign compositions to each each layer. Just porting the algorithm as it was written was extremely inefficient when used in the world builder, so I had to redesign it and also pre-read and compress the data in advance. I was able to get it to run much faster, but it was still not great and also not flexible at all.

This approach is I think a good enough compormise, focusing on what it will probably be used for mostly: determining the depth of features and layers. You can use it in the following way: 
```
"temperature models":[{"model":"linear", "min depth":-1e2, "max depth":"Litho1.0: lithosphere", "top temperature":273.15},
```
or:
```
     "composition models":[
       {"model":"uniform", "min depth":-1e2, "max depth":"Litho1.0: sediment 1", "compositions":[8]},
       {"model":"uniform", "min depth":"Litho1.0: sediment 1", "max depth":"Litho1.0: sediment 2", "compositions":[9]},
       {"model":"uniform", "min depth":"Litho1.0: sediment 2", "max depth":"Litho1.0: sediment 3", "compositions":[10]},
       {"model":"uniform", "min depth":"Litho1.0: sediment 3", "max depth":"Litho1.0: crust 1", "compositions":[11]},
       {"model":"uniform", "min depth":"Litho1.0: crust 1", "max depth":"Litho1.0: crust 2", "compositions":[12]},
       {"model":"uniform", "min depth":"Litho1.0: crust 2", "max depth":"Litho1.0: crust 3", "compositions":[13]},
       {"model":"uniform", "min depth":"Litho1.0: crust 3", "max depth":"Litho1.0: lithosphere", "compositions":[14]}
       ]
``` 

I think this provides resonable flexiablity, and it give a whole new interface for providing other data. So even if you wanted a more complicated litho1.0 model, you could just add a new module with more parameters.

The one thing I am not sure about yet is how to deal with the litho1.0 dataset itself. It is compised of 40962 files with data per location and one file with the all locations totalling more than 500MB. Currently I only need the locations and the depths of the 9 uppermost layers. I could prepack everything I need in plain-text (~3mb), binary (~2mb) or compressed (~500kb), or I could require that users just place the full dataset somewhere themselfs. Since the dataset is (I think) static (it doesn't seem to have been updated since it release in 2014, just new available formats according to the website https://igppweb.ucsd.edu/~gabi/litho1.0.html), I think packing it might be fine if we can get it small enough. Plain text is nice, since people can actually inspect the data and it can be properly managed by git (if there would ever be an update), but it does cost quite a bit of space compared to the compressed version. This might especially start to add up if more depths or other data than just the depth would be needed as well (~2mb per data).  Currently the source code of the world builder as dowloaded for the latest release is ~35mb.

### DIFF
--- a/source/world_builder/features/continental_plate.cc
+++ b/source/world_builder/features/continental_plate.cc
@@ -81,9 +81,9 @@ namespace WorldBuilder
 
       prm.declare_entry("", Types::Object(required_entries), "Continental plate object. Requires properties `model` and `coordinates`.");
 
-      prm.declare_entry("min depth", Types::OneOf(Types::Double(0),Types::Array(Types::ValueAtPoints(0., 2.))),
+      prm.declare_entry("min depth", Types::OneOf(Types::Double(0),Types::Array(Types::ValueAtPoints(0.,2))),
                         "The depth from which this feature is present");
-      prm.declare_entry("max depth", Types::OneOf(Types::Double(std::numeric_limits<double>::max()),Types::Array(Types::ValueAtPoints(std::numeric_limits<double>::max(), 2.))),
+      prm.declare_entry("max depth", Types::OneOf(Types::Double(std::numeric_limits<double>::max()),Types::Array(Types::ValueAtPoints(std::numeric_limits<double>::max(),2))),
                         "The depth to which this feature is present");
       prm.declare_entry("temperature models",
                         Types::PluginSystem("", Features::ContinentalPlateModels::Temperature::Interface::declare_entries, {"model"}),
--- a/source/world_builder/features/continental_plate_models/composition/random.cc
+++ b/source/world_builder/features/continental_plate_models/composition/random.cc
@@ -66,12 +66,12 @@ namespace WorldBuilder
 
           // Declare entries of this plugin
           prm.declare_entry("min depth", Types::OneOf(Types::Double(0),
-                                                      Types::Array(Types::ValueAtPoints(0., 2.)),
+                                                      Types::Array(Types::ValueAtPoints(0.,2)),
                                                       Types::String("")),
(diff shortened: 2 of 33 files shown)


MFraters:
Just as a note, the microsoft/windows/visual studio compiler has a much smaller stacksize than what seems to be the default on linux. Putting the Litho1.0 data in a normal (big) array caused all kind of nasty stack overflow issues which took me a bit to narrow down to and realize what the issue was.

An other option to deal with the Litho1.0 data is to format it as the initialization of a std::vector in a header file and just compile it into the program. The advantage is that no reading of files and checking where those file are and whether they are correct, is neede anymore. This would also make it easier for users since they won't need to worry about there the data is stored on their system. On the dowsides, it would probablyncrease the size of the library binary from 4.5mb to around 6.5mb (havn't tested it how much it would be), and would not be very flexible or scalable to other or updated dataset. One way to mitigate the size would be to use a cmake flag whether to include the data or not, which would enable or disable the relevant plugins.

MFraters:
I am now happy with the implementation, but I still need to do some small cleanup and add a changelog. I went with the approach of directly adding the data into the code through 2 long vectors in a header file. I think this is a resonable compromise for this type of data. The most chanenging part was to get this to work on windows with MSVC, because the (default?) stack size is way smaller than on linux, but it now seems to work on all platforms.

Issue #780, opened 2025-01-18
initial topography


MFraters:
There is already #132, but I think I would like to take it in a slightly different direction. The reason for this is that @Djneu and @alanjyu are interested in initial isostatically equilibrated initial topography. This would make for the for the following requirement list

The user needs to be able to select in the world builder file what kind of topography they want. 
1.  use the highest feature (to match what the world builder prescribes)
2. isostatically equilibrate.

Once we have a good way for the user to input this, the two (or more) different topography outputs can just be models to be chosen. These options can be worked on independently.

-----------------------------

For option 1, we would need to implement the following:

1. create an output property which returns the topography.
2. compute the topography at a certain location by finding the feature with is on top of all the other ones (depth wise) and then find the depth of the top layer in that location.
3. This would probably need on option per feature and/or layer to exclude it from the topography computation. This is because you would not want a sticky air to be counted as topography.

-----------------------

For option 2, we would need to implement the following:

1. A density plugin system, which can return the density at a given point in a feature. This would take as input the location (relative in the feature), absolute depth and temperature (this means that this plugin system is dependent on the temperture, and it won't be allowed to have a dependency the other way around).
2. a intitial topogarphy model which computes an integral over the features above (and below till the compensation depth?) it, calls the gravity and density plugin systems and computes the topography.
3. We need to think about wether to only let pass the topography value and let the calling program (e.g. ASPECT) deal with what to do with the extra height, or that we should be adjusting the location of the features based on the isostatically adjusted topography.

PR #848, opened 2025-08-24
Add topograhpy infrastructure

MFraters:
An alternative implementation to #814 for the topography interface (issue #780). The difference is that this implementation can have different topography models per feature. It is still very much a work in progress, since not all features and geometries are implemented, but is as already workable, as can be seen in the example below. @Djneu would this also work for your use-cases? @alanjyu where you also interested in this, and if so, does this work for you as well? 

I am thining of turning the topography input from a single value into a surface, like the min and max depht. That would allow to seamlessly use Litho1.0 data. (as a side-note, I probably want to rework the current implementation of surfaces to be more flexible and in line with the rest, but that is for a different pull request.)
 
screenshot_topography

This is made with the following world builder file:

```
{
  "version":"1.1",
  "coordinate system":{"model":"spherical", "depth method":"begin at end segment"},
    "surface temperature":293.15,
  "force surface temperature":true,
  "features":
  [
    {"model":"continental plate", "name":"Continent plate A", "coordinates":[[-1,10],[41,10],[41,15],[-1,15]], "min depth":-1e2,
     "topography models":[{"model":"uniform","topography":50000}],
     "temperature models":[{"model":"uniform", "min depth":-1e2, "max depth":95e3, "temperature":500}],
     "composition models":[{"model":"uniform", "compositions":[0], "min depth":-1e2}]},

    {"model":"continental plate", "name":"Continent plate B", "coordinates":[[20,12],[31,12],[31,15],[20,15]], "min depth":-1e2,
     "topography models":[{"model":"uniform","topography":70000}],
     "temperature models":[{"model":"uniform", "min depth":-1e2, "max depth":95e3, "temperature":500}],
     "composition models":[{"model":"uniform", "compositions":[1], "min depth":-1e2}]},

    {"model":"continental plate", "name":"Continent plate C", "coordinates":[[10,12],[20,12],[20,15],[10,15]], "min depth":-1e2,
     "topography models":[{"model":"uniform","topography":-10000}],
     "temperature models":[{"model":"uniform", "min depth":-1e2, "max depth":95e3, "temperature":500}],
     "composition models":[{"model":"uniform", "compositions":[2], "min depth":-1e2}]},

      // no topography and everywhere in the continent. Should not change any topography
    {"model":"continental plate", "name":"Continent plate D", "coordinates":[[-1,10],[41,10],[41,15],[-1,15]], "min depth":-1e2,
     "composition models":[{"model":"uniform", "compositions":[3], "min depth":-1e2, "max depth":25e3}]},

    {"model":"oceanic plate", "name":"oceanic plate A", "coordinates":[[-1,0],[41,0],[41,10],[-1,10]], "min depth":-1e2,
     "temperature models":[{"model":"half space model", "max depth":95e3, "min depth":-1e2, "spreading velocity":0.005, "ridge coordinates":[[[10,0], [10,5]],[[20,5],[20,10]]]}],
     "composition models":[{"model":"uniform", "compositions":[4], "min depth":-1e2}]},

    {"model":"oceanic plate", "name":"oceanic plate B", "coordinates":[[12.5,2.5],[17.5,2.5],[17.5,5],[12.5,5]], "min depth":-1e2,
     "topography models":[{"model":"uniform", "topography":10000}],
     "composition models":[{"model":"uniform", "compositions":[5], "min depth":-1e2}]}
  ]
}
```

### DIFF
--- a/source/world_builder/parameters.cc
+++ b/source/world_builder/parameters.cc
@@ -803,6 +803,11 @@ namespace WorldBuilder
                 for (size_t index = 0; index < n_nodes; ++index )
                   {
                     // only add the point if it is actually in the feature
+                    // TODO?: The performance of models with features which have many Litho1.0 layers (no matter the type), could be theoretically significantly
+                    //        improved by somehow getting this polygon contains point outside the current function and on the composition plugin level or even at
+                    //        the feature level. This is because whether a coordinate is inside or outside a feature doesn't change, so we would only need to do
+                    //        this once per feature. The only problem is that the feature would need to know that litho1.0 is asked for, because otherwise it would
+                    //        be significantly more expensive for model which do not want to use litho1.0.
                     if (Utilities::polygon_contains_point(scaled_polygon, Point<2>(Datasets::LITHO1_0::coordinates_lat_long[index*2+1],Datasets::LITHO1_0::coordinates_lat_long[index*2],CoordinateSystem::spherical)))
                       {
                         const size_t global_index = index*n_layers+static_cast<size_t> (layer_type);


Djneu:
I think this sounds good for now, and we can update it later if it is necessary when isostasy is implemented. From the third point on option 2 in #780, I think the idea was that we pass the topography separately and then ASPECT will decide how to apply the topography, but this was something that we still need to implement within ASPECT for it to work.

PR #858, opened 2025-11-06
Add a few comments to distance_point_from_curved_planes

danieldouglas92:
Just adding a few comments for clarity and removing some comments that seemed unnecessary.


MFraters (approved):
Thanks for the fixes and cleaning up of the comments!

I agree that technically it should be two pull requests, but in this case they are small enough to easily review together and they are separated in two commits, and one is required to get the tests working, so I think it is fine in this case :)

PR #859, opened 2025-11-08
[WIP] Allow for Oblique Slabs/Faults Relative to Surface

danieldouglas92:
Currently, faults and slabs dip from the surface at an angle orthogonal to the surface coordinates. In this PR, I want to add the ability for a user to specify an "obliquity vector" which tells the world builder what direction to dip the feature relative to the surface coordinates. An example of why this is useful can be seen in the images below: 

The convergence of two plates on Earth is rarely perfectly orthogonal to the plate boundary, and in subduction zones there is varying degrees of oblique convergence of a plate relative to the plate margin. Currently, if a user wants to initialize a subducting plate that has undergone oblique convergence, the subducting plate will look like this:
image

The trench is rotated 45 degrees relative to the convergence direction, and so WorldBuilder by default also rotates the slab so that it remains orthogonal to the trench. If I specify an "obliquity vector" that is in the positive x-direction, then the slab looks like this:

image

which maintains the same convergence direction even with an oblique trench.

This is implemented by modifying the `distance_point_from_curved_planes` function in `utilities.cc`, and consists of two main chunks of code:

1. Most of the time the current implementation will use a `bezier_curve` to find a point on the trench/fault that is closest to the check point. In this case, I use the check point and the `obliquity_vector` to create a line, which I then incrementally shift towards the bezier curve. If it intersects the `bezier_curve`, then this intersection point becomes the new "closest point" on the `bezier_curve`. If it doesn't converge, then the "closest point" is set to be a NAN.

2. The `bezier_curve` misses points that are near the terminus of the surface coordinates. The way this is currently handled is definitely not generalized. It won't work for spherical and I don't think it will even work well for more complicated Cartesian geometries. I think this will require modifying the `closest_point_on_curve_segment` function in `bezier_curve.cc`. Right now, my idea is to check what the min/max values of the surface coordinates are, draw a line between the two extrema, and see if the check point (with the obliquity vector) will intersect this line. If it does, I check to see if the intersection point lies within the min/max values.

This is still very much a work in progress, I still need to add the "obliquity vector" as an input parameter that gets passed to `distance_point_from_curved_planes`, but I just wanted to open this PR so that I can get feedback on it! 

@MFraters

### DIFF
--- a/source/world_builder/utilities.cc
+++ b/source/world_builder/utilities.cc
@@ -575,69 +575,6 @@ namespace WorldBuilder
               const Point<2> vector_checkpoint_to_curve = closest_point_on_line_2d - check_point_surface_2d_temp_ob;
               const double line_factor = (obliquity_vector_point * vector_checkpoint_to_curve) >= 0.0 ? 1.0 : -1.0;
 
-              // When trying to uniquely determine which way we should iterate along the obliquity vector (i.e. parallel to it or
-              // anti-parallel to it), which can be seen in the comments above,I was unable to generalize it enough to work for all
-              // cases. Therefore, loop over both cases and quickly see whether one pushes the checkpoint in the wrong directions.
-              const Point<2> initial_iterable_check_point_surface_2d = iterable_check_point_surface_2d;
-              const Objects::ClosestPointOnCurve initial_iterable_closest_point_on_curve = iterable_closest_point_on_curve;
-              // const std::array<double,2> trial_line_factors = {{-1.0, 1.0}};
-              const std::array<double,2> trial_line_factors = {{line_factor}};
-
-
-
-              // bool converged = false;
-              // // Loop over both the parallel and anti-parallel direction of the obliquity vector.
-              // for (const double trial_line_factor : trial_line_factors)
-              //   {
-              //     Point<2> trial_iterable_check_point_surface_2d = initial_iterable_check_point_surface_2d;
-              //     Objects::ClosestPointOnCurve trial_iterable_closest_point_on_curve = initial_iterable_closest_point_on_curve;
-
-              //     // Move the checkpoint along the obliquity vector and search for a converged
-              //     // intersection with the curve.
-              //     for (unsigned int i = 0; i < 100; ++i)
-              //       {
-              //         const double old_dist = std::abs(trial_iterable_closest_point_on_curve.distance);
-
-              //         Point<2> parameterized_line(trial_iterable_check_point_surface_2d[0] + trial_line_factor * old_dist * obliquity_vector_point[0],
-              //                                     trial_iterable_check_point_surface_2d[1] + trial_line_factor * old_dist * obliquity_vector_point[1],
-              //                                     natural_coordinate_system);
-
-              //         trial_iterable_closest_point_on_curve = bezier_curve.closest_point_on_curve_segment(parameterized_line);
-              //         trial_iterable_check_point_surface_2d = parameterized_line;
-              //         const double new_dist = std::abs(trial_iterable_closest_point_on_curve.distance);
-
-              //         const double convergence_tol = bool_cartesian ? 1.0 : 1e-6;
-
-              //         // We are getting farther away, break the loop.
-              //         if (new_dist > old_dist)
-              //           break;
-
-              //         // We converged to a point on the curve, break the loop, break the loop and update the check point.
-              //         if (new_dist < convergence_tol)
-              //           {
-              //             iterable_check_point_surface_2d = trial_iterable_check_point_surface_2d;
-              //             converged = true;
-              //             break;
-              //           }
-              //       }
-
-              //     // If we converged, update the closest point on the bezier curve.
-              //     if (converged)
-              //       {
-              //         closest_point_on_curve   = bezier_curve.closest_point_on_curve_segment(iterable_check_point_surface_2d);
-              //         closest_point_on_line_2d = closest_point_on_curve.point;
-              //         break;
-              //       }
-
-              //     // If we did not converge, assign the closest point as NaN.
-              //     else
-              //       {
-              //         closest_point_on_line_2d[0] = NaN::DQNAN;
-              //         closest_point_on_line_2d[1] = NaN::DQNAN;
-              //       }
-
-              //   }
-
               bool converged = false;
               // Move the checkpoint along the obliquity vector and search for a converged
               // intersection with the curve.


source/world_builder/utilities.cc
```diff
@@ -427,9 +434,166 @@ namespace WorldBuilder
...
+              double new_dist;
+              Point<2> iterable_check_point_surface_2d = check_point_surface_2d;
+              Objects::ClosestPointOnCurve iterable_closest_point_on_curve = bezier_curve.closest_point_on_curve_segment(iterable_check_point_surface_2d);
+              bool converged = false;
+
+              if (reference_normal_on_side_of_line_ob != reference_point_on_side_of_line_ob)
+                {
+                  // We want to take the current checkpoint and move it along the obliquity vector using a parameterized line
+                  // towards the bezier curve until we find a point on the bezier curve that intersects this parameterized line.
+                  for (unsigned int i = 0; i < 100; ++i)
```
MFraters: why 100? What if we don't converge before?

source/world_builder/utilities.cc
```diff
@@ -427,9 +434,166 @@ namespace WorldBuilder
...
+                      // the curve is positive, then we want to move along the obliquity vector in the negative direction,
+                      // otherwise in the positive direction.
+                      line_factor = obliquity_vector * iterable_closest_point_on_curve.normal > 0 ? -1.0 : 1.0;
+
+                      // Parameterize a line through the checkpoint along the obliquity vector.
+                      Point<2> parameterized_line(iterable_check_point_surface_2d[0] + line_factor * old_dist * obliquity_vector[0],
+                                                  iterable_check_point_surface_2d[1] + line_factor * old_dist * obliquity_vector[1],
+                                                  natural_coordinate_system);
+
+                      // Check where the closest point on the bezier curve is to the new check point (after moving along the
```
MFraters:
```suggestion
                      // Check where the closest point on the Bezier curve is relative to the new check point (after moving along the
```
relative to? with respect to?

source/world_builder/utilities.cc
```diff
@@ -427,9 +434,166 @@ namespace WorldBuilder
       double fraction_CPL_P1P2 = std::numeric_limits<double>::infinity();
 
       // Get the closest point on the curve (the trench or the fault trace) to the check point.
-      const Objects::ClosestPointOnCurve closest_point_on_curve = bezier_curve.closest_point_on_curve_segment(check_point_surface_2d);
+      Objects::ClosestPointOnCurve closest_point_on_curve = bezier_curve.closest_point_on_curve_segment(check_point_surface_2d);
       Point<2> closest_point_on_line_2d = closest_point_on_curve.point;
 
+      // Hard-coded obliquity vector for testing purposes, this will be replaced later by an input parameter.
+      // const Point<2> obliquity_vector(1.0, 0.0, natural_coordinate_system);
```
MFraters: note: cleanup when no longer WIP

source/world_builder/utilities.cc
```diff
@@ -427,9 +434,166 @@ namespace WorldBuilder
...
+                      check_point_surface_2d_temp_ob[0]+= 2*Consts::PI;
+                    }
+                  else if (min < normal)
+                    {
+                      check_point_surface_2d_temp_ob[0]-= 2*Consts::PI;
+                    }
+                }
+
+              const Point<2> AB_normal_ob = closest_point_on_curve.normal*closest_point_on_line_2d.distance(reference_point);//*AB.norm();
+              const Point<2> local_reference_point_ob = AB_normal_ob*1.+closest_point_on_line_2d;
```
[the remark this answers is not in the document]
danieldouglas92: This part of the code was copied and I didn't see this! I removed the *1 here and in the other spot as well.

source/world_builder/utilities.cc
danieldouglas92: Yea I don't really have a good reason for picking 100, I was torn between making the number of iterations an input parameter or not. I just chose 100 as a starting number, but if you think it should be higher/a user input parameter I would be ok with changing this.

source/world_builder/utilities.cc
```diff
@@ -406,6 +407,12 @@ namespace WorldBuilder
       const Point<2> check_point_surface_2d(natural_coordinate.get_surface_coordinates(),
                                             natural_coordinate_system);
 
+
+
+
+
+
+
```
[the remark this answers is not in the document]
danieldouglas92: Forgot to remove this, I always add a bunch of empty lines when I'm writing new code to more easily see what I changed

source/world_builder/utilities.cc
danieldouglas92: Removed this!

danieldouglas92:
@MFraters Thanks for taking a look at this so quickly! I addressed the comments on the code you left so far, and for the 4 points you outlined I'll work on addressing in the next couple days. Hopefully you are right that using the end of the bezier curve will work and simplify the code significantly. 

For your points 2-4:

I actually haven't tried doing this in spherical, the current logic (in the loop outside of if `(!std::isnan(closest_point_on_line_2d[0])`) implicitly assumes a cartesian geometry since it just makes a straight line, whereas in spherical it would have to make an arc. If what you suggest works though, then what I have should work in spherical. 

I do agree that this should become part of the sections, this would allow a lot more flexibility in the geometry of the slab along-strike. 

I'm not sure what you mean by the slab seems to extend beyond the plate?

MFraters:
Great, thanks!

> I'm not sure what you mean by the slab seems to extend beyond the plate?

image

danieldouglas92:
Currently, the spherical case is not working. Because I only personally needed the Cartesian case I've been less motivated to fix this, but I'm hoping to have time soon to try to figure out what's going on. This is what the slab looks like for the spherical case when trying to specify that the obliquity vector is to the right:

image

The slab at the very bottom of the trench looks correct (below the tear), but the tear and everything above it is incorrect

MFraters:
Could you describe what you mean with not working robustly? I have some ideas what could potentially cause issues, but it would be good to have a chat I think.

danieldouglas92:
@MFraters Also I messaged you on element, but it would be great to chat about this and also the spherical issue

danieldouglas92:
@MFraters I think I've got something working that is much cleaner. The issue that I discussed with you on zoom for why I couldn't always determine the orientation of the vector is because the distance that was being returned was signed. After I used the absolute value of the distance, the method that I had originally worked.

There is just one thing that is pretty hacky right now that it would be great to get your feedback on. The distance that I'm using to determine convergence is also tricky, if I make it too strict then in Cartesian the iteration won't reliably converge creating gaps in the slab. For spherical though, since the distances returned are not in meters, the tolerance has to be very small. So to get around this, I set the tolerance differently for either cartesian or spherical geometries. Is there a better way to set this tolerance for both geometries? 

I'll write up proper tests and demonstrate that this works with a complicated trench geometry after I get some lunch, and then also add the functionality for faults. But I think this is basically good now.

PR #887, opened 2026-03-26
Doc: add Python instructions

tjhei:


MFraters (approved):
ok, I would say let's merge it for now, since it is an improvement, and make an issue to see if we can fix it later. Thanks again!

PR #895, opened 2026-03-27
Modified typoes in Basic Starter Tutorial 16

Tartaspecio:
Fixed two typoes in BST 16 Mass Conserving


MFraters (approved):
I think I have solved the issue with the equations in #900, so it should be good to merge now. Thanks @Tartaspecio for the extra documentation!

PR #810, opened 2025-01-21
Integrating GPlately for Plate Reconstruction and Feature Extraction

Minerallo:
Hi everyone,

I initially planned to use the GPlates Web Services OpenAPI, as discussed in [issue #787](https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/787). However, I realized that an alternative and potentially more efficient approach would be to leverage [GPlately](https://gplates.github.io/gplately/v1.3.0/)(https://github.com/GPlates/gplately), which already provides a wide range of features such as trenches, ridges, transform faults, continents, cratons, and many more accessible using a python script. GPlately also offers the capability to perform remote plate reconstructions and extract associated geological features seamlessly into geojson format.

**How it work for now**
Using Gplate2gwb.ipynb (Later I will do a .py too), the script runs GPlately to import selected geological features based on a specified plate reconstruction model and geological time (Figure 1). Then we generates multiple GeoJSON files, each corresponding to a specific geological feature. Currently, the script supports the extraction of continents and ridges_and_transform_faults, with plans to include additional features such as trenches, passive margins, cratons etc.
I plan to adapt the script so the user can also decide to import its own local plate reconstruction model as it is currently possible using Geojson2json.py for continental plate.

Geojson2json.py is for now really simple and will probably be combine to Gplate2gwb, it converts GeoJSON files to the World Builder-compatible JSON format.
Using the Gplates GUI the users can locally extract features in geojson format (currently it only support continental plate geometry, and make it directly readable by the Geodynamic World Builder.

**Usage Instructions for Geojson2json.py**
Load a plate reconstruction in GPlates or define your own continental plates.
Export the data to GeoJSON format using the specified parameters (figure 2).
Once the GeoJSON files are generated, run the conversion script using the command:

**Ongoing Work**

- [ ]  Integration of additional geological features: all_topological_sections, all_topologies, coastlines, continent_ocean_boundaries, continental_crusts, continental_rifts, continents, extended_continental_crusts, faults, fracture_zones, inferred_paleo_boundaries, misc_boundaries, misc_transforms, orogenic_belts, passive_continental_boundaries, ridges, ridges_and_transforms, slab_edges, subduction_direction, sutures, terrane_boundaries, transforms, transitional_crusts, trenches, unclassified_features
- [x]  Support for local plate reconstruction in Gplate2gwb.ipynb
- [ ]  Implementation of slab-dip dependent thermal variations and thickness changes
[...]

Issue #833, opened 2025-01-27
Gplates and Gplately integration [On-going Work]


Minerallo:
Hi everyone,

As asked by Menno in the PR https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/810, I moved the Ongoing work list to this issue. 

**Ongoing Work with Gplate and Gplately integration**
- [x]  Support for local plate reconstruction in Gplate2gwb.ipynb
- [ ]  Integration of additional geological features: for now it includes ridges_and_transforms, trenches and continentals, but could include more (all_topological_sections, all_topologies, coastlines, continent_ocean_boundaries, continental_crusts, continental_rifts, continents, extended_continental_crusts, faults, fracture_zones, inferred_paleo_boundaries, misc_boundaries, misc_transforms, orogenic_belts, passive_continental_boundaries, ridges, slab_edges, subduction_direction, sutures, terrane_boundaries, transforms, transitional_crusts, unclassified_features)
- [ ]  Implementation of slab-dip dependent thermal variations and thickness changes
- [ ]  Calculation of continental and oceanic thermal states using seafloor age data
- [ ]  Addition of paleotopography once mesh deformation is there
- [ ] Documentation in the manual
- [ ] In the future, integration of the gplate web server as discuss in the issue https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/787

PR #902, opened 2026-03-27
GPlately2GWB: Global and Regional Model Generation

Minerallo:
## What’s new compared to before (Builds on #810)
GWB26_gplately2gwb

### 🌊 1. Seafloor grid generation (NEW)
- Automatic computation of:
  - Seafloor age grids if needed
  - Spreading rate grids
- Exported as NetCDF and reused in later steps

### 🔧 2. Trench enrichment with geodynamic parameters (NEW)

Each trench is now augmented with:
- Average slab dip (based on @brmather’s Slab-Dip approach: https://github.com/brmather/Slab-Dip)
- Slab thickness
- Trench azimuth
- Dipping point (computed from trench normal)
- Ridge–trench distance
- Spreading velocity (from grids)
- Subducting velocity (from plate kinematics)

👉 Now: trenches carry full kinematic + geometric information for subduction zone thermal and compositional state

### 🧱 3. Subducting slab generation (NEW)
- Trenches can now be converted into:
  - Subducting plate features (GWB slabs) instead of simple faults
- Includes:
  - Segment geometry
  - Dip angle
  - Thickness
  - call for mass-conserving thermal model using kinematic parameters

### 🏔️ 4. Continental thickness & passive margins (NEW)
- Continents are now:
  - Merged into coherent polygons
  - Assigned variable lithospheric thickness
  - Include "passive margin" lithosphere thickness gradient from coast to interior of continents

### 🌍 5. Full model composition (NEW)
- Adds large-scale background structure:
  - Ambient mantle layer
  - Oceanic lithosphere (plate model)
  - Optional LLSVP layer

### 🔄 6. Trench age / velocity correction (NEW)
- Computes trench age from ridge–trench distance
- Replaces spreading velocity with subducting velocity when needed
- Prevents non-physical (negative) ages

### ✂️ 7. Regional model extraction (NEW)
- Clip global models to a bounding box
- Automatically:
  - Trim coordinates
  - Adjust depth-dependent parameters

### 🛠️ 8. Improved GeoJSON → GWB conversion (UPDATED)
- Better handling of:
  - Multi-geometries
  - Depth-dependent fields
  - GWB schema compliance

### 📦 9. Installation & reproducibility (NEW)
- Added installation script:
  - Installation_dependencies_gplatly2gwb.sh

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -38,6 +38,7 @@ extend-exclude = [
     "/doc/world_builder_declarations_open.md", # should be corrected in the original source
     "contrib/catch/*",
     "contrib/world_builder/*",
+    "contrib/gplates/*",
     "*.bib", # bib entries contain many arbitrary abbreviations
     "tests/prmbackslash*", # this test purposefully breaks words on different lines
     ] 


Minerallo:
> @Minerallo : this is really cool, thank you for your contribution! This is similar to what I wanted to do with QGIS files, but cooler ;) . I did not try or read your code, but I can take a look tomorrow evening if no one else gets to it.

Thank you, @alarshi. I wanted to push it so others can also start adding new features. It is important to keep in mind that this is still a work in progress, because I had to apply several corrections to make it work. For example, there were cases where the velocities provided by GPlates became too high, or where the oceanic age at trenches was somehow lower than 0. So there is still work to be done in the future to refine and adjust these functions. I will add the remaining tasks to the open gplate issue: #833.

Minerallo:
> @Minerallo: Sounds good, let me know when you want me to take a look at it. I would be in the favor of merging this if the features you already implemented are working, even if all the features are not added in this hack.

Thanks @alarshi, I agree, there’s definitely room for improvement in the future, at least this provides a basis. I’ve rebased GWB onto main, and I’ll quickly add an option to export paleotopography/LAB and crustal thickness back into the .wb file before pushing an updated version.

PR #903, opened 2026-03-28
along interface velocity for subducting slab

lhy11009:
Add an along surface velocity option for the subducting slab.  This will trace the surface of the subducting slab and rotate the velocity to align with the surface. The free parameter is the magnitude of the velocity.

Here is the layout of the test, which connects a horizontal velocity in the plate to the along-surface velocity in the slab.
image

This is tentative. Eventually, we might want to tweak the velocity in each of the slab segments, but this would be an easy start.


MFraters:
Cool! Yes, the next step would be have a velocity per segment something like the angles, either [1.5] or [1.5,2.2].

lhy11009:
If we add a sign modification to Vx depending on the relative position of the dip point and the trench point, and if dip point have smaller x than the trench point, we change the sign of Vx. 
With Menno's help, I also adjusted the angle I use to be the local angle; now the velocity perfectly aligns with the surface.
This angle is added to the distance_from_plane object

image

source/world_builder/features/subducting_plate_models/velocity/along_surface.cc
```diff
@@ -0,0 +1,122 @@
...
+                                   const double  /*feature_max_depth*/,
+                                   const WorldBuilder::Utilities::PointDistanceFromCurvedPlanes &distance_from_plane,
+                                   const AdditionalParameters & /*additional_parameters*/) const
+        {
+
+          if (distance_from_plane.distance_from_plane <= max_depth && distance_from_plane.distance_from_plane >= min_depth)
+            {
+              const double angle = distance_from_plane.angle;
+
+              const double sign_vx = -1.0; // place holder for modify vx direction
```
lhy11009: @MFraters. Now the question is how to figure out when sign_vx should be 1.0 versus -1.0. -1.0 is required to fix the Vx when the query point is left to the trench point with a smaller X coordinate. I wonder this could be figured out from the distance_from_plane object, or we have to pass the reference point to this function.

MFraters: Yes, I think `reference_on_side_of_line` is the variable you are looking for to add to the structure.

lhy11009:
reference_on_side_of_line doesn't seem to give me the structure I want. I return it in the "distance_from_curved_plane" function and multiply it to the x component:

As the result, I can get one test right:
image

But not the other one, which has a slab dipping in the opposite direction:
image

MFraters:
hmm, can you take the `closest_trench_point` from `PointDistanceFromCurvedPlanes` to compute a direction vector?

lhy11009:
Yes, I think that works. Now the two tests are merged into one, where two slabs dip in different directions, and the velocity is along the surface. of the features. I think this fulfills our purpose here.
image

lhy11009:
@MFraters can you take another look at this? I think all tests have passed.

MFraters (commented):
Very cool! Thanks for working on this. 

The code looks good to me. Have  you also tested  this in 3D and in spherical, to make sure that also works as expected (and maybe add tests for that as well)?

lhy11009:
@MFraters , I follow the cookbook of 3d_spherical_subduction, and add a velocity of 1.0 with the "along surface" model.
In the figure below, I plot the contour of temperature as well as the velocity vectors.
The slab dip decreases to 0 at the end, but notice that the vector doesn't track this change:

image

source/world_builder/features/subducting_plate_models/velocity/along_surface.cc
```diff
@@ -0,0 +1,124 @@
...
+                                   std::array<double,3> velocity_,
+                                   const double  /*feature_min_depth*/,
+                                   const double  /*feature_max_depth*/,
+                                   const WorldBuilder::Utilities::PointDistanceFromCurvedPlanes &distance_from_plane,
+                                   const AdditionalParameters & /*additional_parameters*/) const
+        {
+
+          if (distance_from_plane.distance_from_plane <= max_depth && distance_from_plane.distance_from_plane >= min_depth)
+            {
+              const double angle = distance_from_plane.angle;
```
lhy11009, line 99: It seems this angle I use is not tracking the angle of the slab curvature in this case, which worked fine in 2-D test.

MFraters:
Can you test 3D cartesian, 2D spherical and 3D spherical seperately? That might make it easier to track down whether the issue is 3D or spherical coordinates.

lhy11009:
I seem to come across another bug testing the Cartesian model. When I take the 3d_cartesian_curved_subduction cookbook and run the model, initially it ran through and produced a vtu outputs.

But when I add the "along surface" velocity to the subducting plate feature, I received an Arithmetic exception.
I output the variables in gdb and the "angle" I computed is "nan" value, and the point this happens (650000, 1000000, 1000000).

Here is the layout of the point (white point) with a 1473 K iso-volume plot of the slab.
image
, and from another angle:
image

source/world_builder/utilities.cc
```diff
@@ -867,6 +869,7 @@ namespace WorldBuilder
                           new_distance = side_of_line * (check_point_2d - Pb).norm();
                           new_along_plane_distance = (begin_segment - Pb).norm();
                           new_depth_reference_surface = start_radius - Pb[1];
+                          new_check_point_angle = interpolated_angle_top;
```
lhy11009, line 872: The "angle" ultimately comes from this "interpolated_angle_top)

source/world_builder/features/subducting_plate_models/velocity/along_surface.cc
```diff
@@ -0,0 +1,134 @@
...
+              // WBAssertThrow(std::isfinite(angle), "Invalid angle value: angle = " << angle);
+
+              // determine the value of vx
+              const double sign_vx = (position_in_cartesian_coordinates[0] > distance_from_plane.closest_trench_point[0])? 1.0: -1.0;
+
+              // todo_velo
+              double vx = 0.0;
+              double vy = 0.0;
+              double vz = 0.0;
+              if (std::isfinite(angle))
```
lhy11009, line 111: I tried to bypass this error if the error is related to a few points with this added line.

lhy11009: @MFraters , But then I got 0 velocity everywhere, meaning the angle is nan everywhere for this case. Is it this wb have multiple sections in the slab, and that causes the problem? I can try a different test file later, but maybe you could first comment on this issue.

MFraters:
I read through the comments, and it looks like we need to take a look at when and why this angle is actually computed. Since I am at a conference right now I don't have the time to really look deep into the code right now, but I will try to do this after the conference.

lhy11009:
@MFraters, would you have a time next week to look at this? Perhaps we could get this and my other PR merged before the Hackathon.

MFraters:
Sorry it took me a bit to get around to this. The issue with the segfault has to do with an early return case around line 615 in utilities.cc. You need to add `return_values.angle = total_average_angle;` there.

Your current approach to get the directly only works in 2D though. I think you should export the `x_axis` variable (around line 684). I think that might get you the right direction both in Cartesian and spherical.
 
image

lhy11009:
@MFraters Thanks for your reply. I would need to get back to this after the ASPECT Hack.

lhy11009:
Good, I seem to have fixed the problem, at least for Cartesian geometry. I also added a new angle to the `PointDistanceFromCurvedPlanes` struct that tracks the direction of the `x_axis`. This allows us to predict the velocity direction in 3D as well. Below are test results in 2-d and in 3-d.

image
image

I will then test the implementation in Spherical geometry. 

@MFraters I have to get to the bottom of the distance_point_from_curved_planes function to reach this point. This function is indeed a very thoughtful implementation and really helps me understand how GWB works in general. I do still have a few small questions to ask, will you have a 30 mins meeting time next week?

lhy11009:
Here are the results from the 3-D spherical test:

image

To achieve this, I added two additional functions to the utilities. I found that the utilities already contain coordinate transformations between Cartesian and spherical geometries, but seem to be missing the corresponding vector transformations.

These new functions transform a vector between its local `(r, north, east)` components and its global Cartesian `(x, y, z)` components at a given position in spherical geometry. The implementation is essentially a matrix multiplication using the local spherical basis vectors.

I think these functions could also be generally useful beyond this particular application.

Why are these functions needed here?

The first step is to determine the dipping angle and the azimuth angle, where the azimuth angle is derived from the `a_axis`. The natural way to define an azimuth angle is as an angle on a surface of constant radius. In Cartesian geometry, this is simply the `x-y` plane, while in spherical geometry it corresponds to the local north-east (`n-e`) plane. This is where the transformation from a Cartesian vector to its local spherical components is needed.

When computing the velocity, it is also natural to first express it in the local coordinate system. In Cartesian geometry, these are simply `(Vx, Vy, Vz)`, while in spherical geometry they are `(Vr, Vn, Ve)`. Therefore, after computing the velocity in the local spherical coordinate system, we need the inverse transformation from the local spherical vector back to the global Cartesian `(x, y, z)` vector to obtain the final velocity.

MFraters:
very cool pictures! Yes, lets discuss.

source/world_builder/utilities.cc
```diff
@@ -610,6 +697,8 @@ namespace WorldBuilder
                   return_values.fraction_of_segment = 0.0;
                   return_values.section = i_section_min_distance;
                   return_values.segment = 0;
+                  return_values.angle_x_axis = new_check_point_angle;
```
MFraters: shouldn't this be new_check_point_angle_x_axis?

include/world_builder/features/subducting_plate_models/velocity/along_surface.h
```diff
@@ -0,0 +1,96 @@
...
+                                              const double gravity,
+                                              std::array<double,3> velocity,
+                                              const double feature_min_depth,
+                                              const double feature_max_depth,
+                                              const WorldBuilder::Utilities::PointDistanceFromCurvedPlanes &distance_from_planes,
+                                              const AdditionalParameters &additional_parameters) const override final;
+
+
+          private:
+            // uniform raw velocity submodule parameters
```
MFraters: not that important, but incorrect name. I would just remove this line.

source/world_builder/features/subducting_plate.cc
```diff
@@ -801,7 +801,9 @@ namespace WorldBuilder
 
                             for (const auto &velocity_model: segment_vector[current_section][current_segment].velocity_systems)
                               {
+                                WBAssert(std::isfinite(distance_from_planes.angle), "distance_from_planes is not a infinite number: " << distance_from_planes.angle << ".");
```
MFraters: not a finite number, instead of infinite number.

MFraters (commented):
1. Could you also add your 2D tests to the tester, so we make sure we catch regressions?
2. while reviewing myself, I asked both claude and chatgpt for a review and they found an interesting issue. I looked into it and I do think it is a real bug. If the point is exactly on the line (`closest_point_on_line_cartesian - check_point_surface_cartesian`=0), atan2 has no meaning:
```c++
          Point<3> y_axis = closest_point_on_line_cartesian - closest_point_on_line_bottom_cartesian;
          Point<3> x_axis = closest_point_on_line_cartesian - check_point_surface_cartesian;

          // Calculate the azimuth angle of the local x-axis
          if (natural_coordinate_system == CoordinateSystem::cartesian)
            {
              new_check_point_angle_x_axis = std::atan2(x_axis[1], x_axis[0]);
            }
          else
            {
              std::array<double,3> x_axis_spherical = cartesian_vector_to_local_spherical(check_point_surface_2d_array,
              std::array<double, 3> {{x_axis[0], x_axis[1], x_axis[2]}});
              new_check_point_angle_x_axis = std::atan2(x_axis_spherical[2], x_axis_spherical[1]);
            }

          // This are accounting for corner cases.
          // If the point to check is exactly on or below the line, we can not compute the x-axis with this method.
          // We could use an other method where we use the two point before and after it, but we can also
          // just nudge it into a direction, which seems to work very well.
          if (std::fabs((check_point_surface - closest_point_on_line_surface).norm()) < 2e-14)
            {
              // The surface coordinates are very close to the line, but the depth coordinate is not. The point is beneath the line.
              if (std::fabs((check_point - closest_point_on_line_cartesian).norm()) > 2e-14)
                {
                  const Point<2> &P1(point_list[i_section_min_distance]);
                  const Point<2> &P2(point_list[i_section_min_distance+1]);
                  ...
                  x_axis = Point<3>(ux*ux*vx + ux*uy*vy - uz*vy + uy*uz*vz + uy*vz,
                                    uy*ux*vx + uz*vx + uy*uy*vy + uy*uz*vz - ux*vz,
                                    uz*ux*vx - uy*vx + uz*uy*vy + ux*vy + uz*uz*vz,
                                    cartesian);
                  ...
                  x_axis = x_axis *(reference_on_side_of_line / x_axis.norm());

                  WBAssert(!std::isnan(x_axis[0]),
                           "Internal error: The x_axis variable is not a number: " << x_axis[0]);
                  WBAssert(!std::isnan(x_axis[1]),
                           "Internal error: The x_axis variable is not a number: " << x_axis[1]);
                  WBAssert(!std::isnan(x_axis[2]),
                           "Internal error: The x_axis variable is not a number: " << x_axis[2]);
                }
              else

                // The point to check is on the line, we don't need to search any further, because we know the distance is zero.
                {
                  total_average_angle = plane_segment_angles[original_current_section][0][0]
                                        + fraction_CPL_P1P2 * (plane_segment_angles[original_next_section][0][0]
                                                               - plane_segment_angles[original_current_section][0][0]);

                  PointDistanceFromCurvedPlanes return_values(natural_coordinate.get_coordinate_system());
                  return_values.distance_from_plane = 0.0;
                  return_values.distance_along_plane = 0.0;
                  return_values.fraction_of_section = fraction_CPL_P1P2;
                  return_values.fraction_of_segment = 0.0;
                  return_values.section = i_section_min_distance;
                  return_values.segment = 0;
                  return_values.angle_x_axis = new_check_point_angle;
                  return_values.angle = total_average_angle;
                  return_values.average_angle = total_average_angle;
                  return_values.depth_reference_surface = 0.0;
                  return_values.closest_trench_point = closest_point_on_line_cartesian;
                  return return_values;
                }
            }

          // The point is not on or beneath the line.
          else
            {
            ...
```

If closest_point_on_line_cartesian and check_point_surface_cartesian are equal, then `x_axis = closest_point_on_line_cartesian - check_point_surface_cartesian;` is zero. This means that `new_check_point_angle_x_axis = std::atan2(x_axis[1], x_axis[0]);` is atan2(0,0) and the setup is not design to handle this (the x_axis vector has no direction). 

I think there are two possible fixes. The most thorough would be to rewrite this to just use and return vectors instead of angles, but that would probably be quite a bit of work. I think the easiest fix is to do a small reorganization of this code, something like this:
```c++
                    Point<3> y_axis = closest_point_on_line_cartesian - closest_point_on_line_bottom_cartesian;
          Point<3> x_axis = closest_point_on_line_cartesian - check_point_surface_cartesian;

          // This are accounting for corner cases.
          // If the point to check is exactly on or below the line, we can not compute the x-axis with this method.
          // We could use an other method where we use the two point before and after it, but we can also
          // just nudge it into a direction, which seems to work very well.
          if (std::fabs((check_point_surface - closest_point_on_line_surface).norm()) < 2e-14)
            {

                  const Point<2> &P1(point_list[i_section_min_distance]);
                  const Point<2> &P2(point_list[i_section_min_distance+1]);
                  ...
                  x_axis = Point<3>(ux*ux*vx + ux*uy*vy - uz*vy + uy*uz*vz + uy*vz,
                                    uy*ux*vx + uz*vx + uy*uy*vy + uy*uz*vz - ux*vz,
                                    uz*ux*vx - uy*vx + uz*uy*vy + ux*vy + uz*uz*vz,
                                    cartesian);
                  ...
                  x_axis = x_axis *(reference_on_side_of_line / x_axis.norm());

                  WBAssert(!std::isnan(x_axis[0]),
                           "Internal error: The x_axis variable is not a number: " << x_axis[0]);
                  WBAssert(!std::isnan(x_axis[1]),
                           "Internal error: The x_axis variable is not a number: " << x_axis[1]);
                  WBAssert(!std::isnan(x_axis[2]),
                           "Internal error: The x_axis variable is not a number: " << x_axis[2]);

                  // Calculate the azimuth angle of the local x-axis
                  if (natural_coordinate_system == CoordinateSystem::cartesian)
                  {
                    new_check_point_angle_x_axis = std::atan2(x_axis[1], x_axis[0]);
                  }
                  else
                  {
                    std::array<double,3> x_axis_spherical = cartesian_vector_to_local_spherical(check_point_surface_2d_array,
                    std::array<double, 3> {{x_axis[0], x_axis[1], x_axis[2]}});
                    new_check_point_angle_x_axis = std::atan2(x_axis_spherical[2], x_axis_spherical[1]);
                  }
            
              // The point to check is on the line, we don't need to search any further, because we know the distance is zero.
              if (std::fabs((check_point - closest_point_on_line_cartesian).norm()) <= 2e-14)
                {
                  total_average_angle = plane_segment_angles[original_current_section][0][0]
                                        + fraction_CPL_P1P2 * (plane_segment_angles[original_next_section][0][0]
                                                               - plane_segment_angles[original_current_section][0][0]);

                  PointDistanceFromCurvedPlanes return_values(natural_coordinate.get_coordinate_system());
                  return_values.distance_from_plane = 0.0;
                  return_values.distance_along_plane = 0.0;
                  return_values.fraction_of_section = fraction_CPL_P1P2;
                  return_values.fraction_of_segment = 0.0;
                  return_values.section = i_section_min_distance;
                  return_values.segment = 0;
                  return_values.angle_x_axis = new_check_point_angle_x_axis;
                  return_values.angle = total_average_angle;
                  return_values.average_angle = total_average_angle;
                  return_values.depth_reference_surface = 0.0;
                  return_values.closest_trench_point = closest_point_on_line_cartesian;
                  return return_values;
                }
            }

          // The point is not on or beneath the line.
          else
            {
            ...
```
also, I changed `return_values.angle_x_axis = new_check_point_angle; ` to `return_values.angle_x_axis = new_check_point_angle_x_axis;` . I think that means the two atan2 computations can be merged into one computation after the x_axis is known (I also adjusted that). 

Although chatgpt says that atan2 computation need to be after `x_axis = Point<3>(ux*ux*vx ...` and  before ` x_axis = x_axis *(reference_on_side_of_line / x_axis.norm());`. I need to think a bit more about that before giving an opinion on it, but maybe you can have a look and let me know what you think.

lhy11009:
The 2-d test is already there, it's the one without the "3D" in its name.

lhy11009:
Hi @MFraters, So what question 2 really is is this:

What if the check_point and closest_ppint_on_line_surface is the same?
This is indeed an issue.

However, I think you already have this handled in the function. While my problem is I seem to place this at the wrong line.

source/world_builder/utilities.cc
lhy11009, line 576: Look at this line here, you already handled in below.

lhy11009, line 698: I think what I should do is move the calculation of new_check_point_angle_x_axis outside of the if-else block above into this else block.

source/world_builder/utilities.cc
lhy11009: Yes, thanks for catching that.

include/world_builder/features/subducting_plate_models/velocity/along_surface.h
lhy11009: Removed.

source/world_builder/features/subducting_plate.cc
lhy11009: Yes, corrected.

source/world_builder/utilities.cc
lhy11009: I did that, see if this makes sense.

lhy11009:
Typo check is failing. It doesn't seem like I typed these "latitutde" instances, should I just fix them?

MFraters (commented):
1. `The 2-d test is already there, it's the one without the "3D" in its name.` Yes, you are right, but it is only the cartesian case in 2D. Since you already have the 2D spherical case (you had an image of it at least), it would be great to add that as well. 
2. `Typo check is failing. It doesn't seem like I typed these "latitutde" instances, should I just fix them?` No, lets leave that for a different pull request. If it was not introduced in this pull request lets not mix it up. This has to do that the typos program updates automatically in the tester. I think in aspect it is fixed, but I don't think for the world builder it is an issue if new problems are discovered (even though they are not related to the pull request.
3. I think your fix only partly solves the problem. Currently if `std::fabs((check_point - closest_point_on_line_cartesian).norm())` is smaller than or equal to `2e-14` (so `if (std::fabs((check_point - closest_point_on_line_cartesian).norm()) > 2e-14)` is false, then the new_check_point_angle_x_axis you return is 0 (that is what it is initialzed to), not an actually reasonable orthogonal direction to the trench. (I don't know if we actually want it to ever return an initialized value, maybe we should initialize it as NaN) The code below the `if (std::fabs((check_point - closest_point_on_line_cartesian).norm()) > 2e-14)` actually checks for this and handles the special case with the line: `const Point<2> closest_point_on_line_plus_normal_to_plane_spherical = closest_point_on_line_2d + 1e-8 * (closest_point_on_line_2d.norm() > 1.0 ? closest_point_on_line_2d.norm() : 1.0) * unit_normal_to_plane_spherical;`, but that is not run due to the if statement being above this statement. So I think it only requires a small reorganization of the if statements as suggested in my earlier comment.

lhy11009:
New figure for 3-d test after update point 3:

image

lhy11009:
The 2-D spherical case has two separate bugs:

image

The first bug is that the computed \(y\)-velocity does not appear to be written correctly to the VTU file, even though the GDB output shows large values.

image

The second, smaller bug is that, in 2-D, the calculated `angle_x_axis` is slightly different from \(-\pi/2\). As a result, part of the velocity is assigned to `Vn`.

Do you have any idea?

lhy11009:
- For the first problem, there is now a change in the world.cc:  

This changes how the vector is projected from 3-D to 2-D

Cartesian geometry still receives exactly the same projection. Spherical geometry skips it and preserves the complete three-component Cartesian vector because its local radial direction changes with position and cannot be represented by assuming that global z is always vertical.

I'm not sure this is the final fix, but I think the current projection doesn't for spherical, at least.

- For the second problem, the angle_x_axis is not -pi/2 because:

 The trench is a curved Bézier path, so its local dip direction at the query point is not exactly eastward. That explains the −92.18° azimuth

The caveat of this is the magnitude is about 2% smaller than the given magnitude

lhy11009:
Now the test looks correct:

image

MFraters:
This looks good to me, thanks for the work! Does it now all work/is this ready for another review?

source/world_builder/world.cc
```diff
@@ -463,12 +463,14 @@ namespace WorldBuilder
...
             {
-              // convert 3d velocity vector to a 2d one
-              Point<2> vector = Point<2>(cartesian);
-              vector[0] = surface_coord_conversions[0]*results[counter]+surface_coord_conversions[1]*results[counter+1];
-              vector[1] = results[counter+2];
-              results[counter] = surface_coord_conversions[0]*results[counter]+surface_coord_conversions[1]*results[counter+1];
-              results[counter+1] = results[counter+2];
+              // Cartesian cross sections return the in-plane components. Spherical
+              // cross sections retain the in-plane Cartesian vector returned by the 3d query.
+              if (coordinate_system == cartesian)
```
lhy11009, line 468: Here is the change for projection

lhy11009:
@MFraters  Yes, this is ready for another review. 

Do help me check that the change in 2-D projection is designated for spherical geometry.

BTW, I found that using AI agent to create unit tests is a helpful workflow. When something fails, this could diagnose where it fails and separate potential problems.

source/world_builder/utilities.cc
```diff
@@ -512,91 +576,99 @@ namespace WorldBuilder
           if (std::fabs((check_point_surface - closest_point_on_line_surface).norm()) < 2e-14)
             {
               // The surface coordinates are very close to the line, but the depth coordinate is not. The point is beneath the line.
```
MFraters: I think this line should be (re)moved now that the line below has been removed/moved.

tests/unit_tests/unit_test_world_builder.cc
```diff
@@ -2711,6 +2712,227 @@ TEST_CASE("WorldBuilder Features: Subducting Plate")
...
+  Utilities::PointDistanceFromCurvedPlanes distance_from_plane(spherical);
+  distance_from_plane.distance_from_plane = 0.0;
+  distance_from_plane.angle = 45.0 * Consts::PI / 180.0;
+  distance_from_plane.angle_x_axis = 0.5 * Consts::PI;
+
+  const std::array<double,3> velocity = velocity_model.get_velocity(position,
+                                                                    natural_coordinate,
+                                                                    depth,
+                                                                    0.0,
+  {{0.0,0.0,0.0}},
```
MFraters, line 2784: hmm, interesting that astyle doesn't recognize of fix this. Anyway, that is not holding back this pull request, just a note for me to investigate.

MFraters (approved):
This looks great now, thanks for all the work and sticking through the review! Just one small comment (and one comment for me you can ignore). Can you also add a changelog entry? Otherwise, I think it is good to merge :)

source/world_builder/features/subducting_plate_models/velocity/along_surface.cc
```diff
@@ -0,0 +1,144 @@
...
+        AlongSurface::declare_entries(Parameters &prm, const std::string & /*unused*/)
+        {
+          // Document plugin and require entries if needed.
+          // Add `velocity magnitude` and to the required parameters.
+          prm.declare_entry("", Types::Object({"velocity magnitude"}),
+                            "Uniform velocity model. Set the velocity to a constant value.");
+
+          // Declare entries of this plugin
+          prm.declare_entry("min distance slab top", Types::Double(0),
+                            "todo The depth in meters from which the composition of this feature is present.");
```
MFraters, line 65: Can you do this todo in a follow up pull request?

source/world_builder/utilities.cc
```diff
@@ -444,8 +505,11 @@ namespace WorldBuilder
       // The 'vertical' fraction, indicates how far in the current segment the
       // point on the line is.
       double segment_fraction = 0.0;
+      double current_angle = 0.0;
       double total_average_angle = 0.0;
       double depth_reference_surface = 0.0;
+      double new_check_point_angle = 0.0;
+      double new_check_point_angle_x_axis = 0.0;
```
MFraters, line 512: I think this should be NaN. Can you do this in a follow up pull request?

MFraters (approved):
Thanks looks good, but I have two small comments. 

I don't want to further hold back this pull request, so I will merge it as is, but I would like to ask you to make two small follow up pull request to make some small local fixes.

Issue #893, opened 2026-03-27
Initial textures from input file and second-order structure tensor


Wang-yijun:
I'm planning to implement two new options to have initial textures on particles. The first one is to read from a standard input file (e.g. ASCII file like the particle cpo output from ASPECT), and the second one is to sample an ensemble of Euler angles from a set of second-order structure tensors as input. An example of the second method implemented with Python can be found here: https://github.com/Theo-Geo/cpo-sampling

PR #905, opened 2026-03-28
Option to read initial texture from input file

Wang-yijun:
I added the option to read the initial texture from an input file as a grain model for all features. Something to note about this grain model:
- It will check that the number of particles and the number of grains in the input file are the same as provided by the ASPECT input file. 
- If there are multiple features, it allows each feature to have different initial textures (from an input file or using the other grain models), and the number of particles within this feature also needs to match the number of particles inside the input file provided. - Both Olivine and Enstatite are allowed. 
- It expects the input file to have the format of a particle_cpo output file from ASPECT. It expects an input file with headers (Olivine only):
  - id, mineral_0_volume_fraction, mineral_0_EA_phi, mineral_0_EA_theta, mineral_0_EA_z) 
- or headers (Olivine and Enstatite):
  - id mineral_0_volume_fraction mineral_0_EA_phi mineral_0_EA_theta mineral_0_EA_z mineral_1_volume_fraction mineral_1_EA_phi mineral_1_EA_theta mineral_1_EA_z 

I also added the random_uniform_distribution grain model for the plume feature. This was missing when I last worked on this.

This is the first part mentioned in https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/893

### DIFF
--- a/include/world_builder/features/plume_models/grains/ascii_file.h
+++ b/include/world_builder/features/plume_models/grains/ascii_file.h
@@ -97,6 +97,7 @@ namespace WorldBuilder
           private:
 
 
+            bool file_loaded;
             std::map<unsigned int, std::vector<std::array<double,3>>> mineral_0_euler_angles_map;
             std::map<unsigned int, std::vector<std::array<double,3>>> mineral_1_euler_angles_map;
             std::map<unsigned int, std::vector<double>> mineral_0_volume_fraction_map;
--- a/source/world_builder/features/mantle_layer_models/grains/ascii_file.cc
+++ b/source/world_builder/features/mantle_layer_models/grains/ascii_file.cc
@@ -48,6 +48,7 @@ namespace WorldBuilder
       {
         AsciiFile::AsciiFile(WorldBuilder::World *world_)
           :
+          file_loaded(false),
           min_depth(NaN::DSNAN),
           max_depth(NaN::DSNAN),
           mineral_0_current_particle_index(0),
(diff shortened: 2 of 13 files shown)


include/world_builder/features/plume_models/grains/ascii_file.h
```diff
@@ -0,0 +1,128 @@
...
+
+            /**
+             * declare and read in the world builder file into the parameters
+             * class
+             */
+            void parse_entries(Parameters &prm) override final;
+
+            /**
+             * Returns a grains based on the given position, composition (e.g.
+             * olivine and/or enstatite)depth in the model, gravity and current grains.
```
Djneu:
```suggestion
             * olivine and/or enstatite) depth in the model, gravity and current grains.
```

source/world_builder/features/mantle_layer_models/grains/ascii_file.cc
```diff
@@ -0,0 +1,279 @@
...
+          min_depth_surface = Objects::Surface(prm.get("min depth",coordinates));
+          min_depth = min_depth_surface.minimum;
+          max_depth_surface = Objects::Surface(prm.get("max depth",coordinates));
+          max_depth = max_depth_surface.maximum;
+          compositions = prm.get_vector<unsigned int>("compositions");
+
+          directory   = prm.get<std::string>("file directory");
+          filename    = prm.get<std::string>("filename");
+
+          load_file();
```
MFraters: Can you try adding a check if the filename is empty, then not call load_file? Also add an assert in get grains, whether the load files was called. I think that should fix the testers.

PR #908, opened 2026-03-28
oceanic plate feature with half space cooled topography model

simransuresh:
- half space cooling temperature model with age as outcome (Turcotte and Schubert book)
- age plugged into the initial topography of the model.


danieldouglas92 (commented):
Hey @simransuresh 

Thanks for this, as I mentioned already I'm excited to try this feature in my own subduction models. I only had two minor comments on your code

danieldouglas92 (commented):
Thanks for addressing my comments @simransuresh! Sorry I missed this before, but please change the name of the class to follow the camel case standard used elsewhere in the world builder. Also, did you see my previous comment about plotting the topography against the solution for subsidence with age (i.e. from Turcotte & Schubert)?

simransuresh:
> Thanks for addressing my comments @simransuresh! Sorry I missed this before, but please change the name of the class to follow the camel case standard used elsewhere in the world builder. Also, did you see my previous comment about plotting the topography against the solution for subsidence with age (i.e. from Turcotte & Schubert)?

image

sorry I missed to add it! Here it is, I hope this is what you asked for - topo vs plate age

simransuresh:
Even with "make indent", my tests are still failing :( I am trying to debug..

MFraters:
you checked in your build directory. please remove it from your commits. That that should fix the indenter

source/world_builder/features/oceanic_plate_models/topography/half_space_cooling.cc
```diff
@@ -0,0 +1,193 @@
...
+                                                   ridge_migration_times);
+          double ridge_distance = ridge_parameters[1];
+          double spreading_velocity = ridge_parameters[0];
+          double age = ridge_distance / spreading_velocity; // in sec
+
+          const double thermal_diffusivity = this->world->thermal_diffusivity;
+          const double alpha = this->world->thermal_expansion_coefficient;
+
+          // age of half space cooling model is used to compute the heights and added with initial depth and topo
+          double half_space_cooling_height = (2.0 * bottom_density * alpha * (bottom_temperature - top_temperature) / top_density) *
```
alarshi: Could you point me to the resource where you get this? This is what I am looking at from Turcotte and Schubert:

Image

which means that the denominator you have looks different or maybe you have a different interpretation for top_density?

simransuresh: Yes you're right, this is the formula i referred to (first commit in this PR). But when adding the top density as input param I missed to add the "bottom density -" too in the denom. Apology for the that. I will recommit with the correct formula.

simransuresh:
@alarshi and @danieldouglas92 I have addressed all your comments.

source/world_builder/features/oceanic_plate_models/topography/half_space_cooling.cc
```diff
@@ -0,0 +1,194 @@
...
+        void
+        HalfSpaceCooling::declare_entries(Parameters &prm, const std::string & /*unused*/)
+        {
+          // Document plugin and require entries if needed.
+          // Add `topography` and half space model params to the required parameters.
+
+          prm.declare_entry("", Types::Object({"ridge coordinates", "spreading velocity",
+                                               "max depth", "min ocean depth", "bottom density",
+                                               "top density"
+                                              }), "Half space cooled topography");
```
danieldouglas92, line 68:
```suggestion
                                              }), "Half space cooling topography");
```

danieldouglas92 (commented):
@simransuresh If the new formula you added reproduces the expected topography from turcotte & schubert then I think your code looks good to me! Just one quick fix with changing "cooled" to "cooling" in the doc string for the class.

Three other quick things:

1) Don't forget to add an entry to the changelog!
2) Please remove world_builder_declarations.tex from your committed files
3) Please squash your commits down to just a couple.

If @alarshi agrees, once you've done this I think this is good to merge! Thanks again for adding this I think this is really great

source/world_builder/features/oceanic_plate_models/topography/half_space_cooling.cc
alarshi: No need for apology, we all make mistakes and that's what reviews are for ;)

PR #915, opened 2026-03-30
fix typo for grains models

Wang-yijun:
This is a small fix for white space and typo in a couple of the grains models.


MFraters (approved):
Great, thanks!

PR #918, opened 2026-03-30
Option to read initial texture from ascii file

Wang-yijun:
I added the option to read the initial texture from an input file as a grain model for all features. Something to note about this grain model:

It will check that the number of particles and the number of grains in the input file are the same as provided by the ASPECT input file.
If there are multiple features, it allows each feature to have different initial textures (from an input file or using the other grain models), and the number of particles within this feature also needs to match the number of particles inside the input file provided. - Both Olivine and Enstatite are allowed.
It expects the input file to have the format of a particle_cpo output file from ASPECT. It expects an input file with headers (Olivine only):
id, mineral_0_volume_fraction, mineral_0_EA_phi, mineral_0_EA_theta, mineral_0_EA_z)
or headers (Olivine and Enstatite):
id mineral_0_volume_fraction mineral_0_EA_phi mineral_0_EA_theta mineral_0_EA_z mineral_1_volume_fraction mineral_1_EA_phi mineral_1_EA_theta mineral_1_EA_z


danieldouglas92:
This looks great, thanks for dividing up your original PR into these small PR's! The coverage has dropped by several percent because your test only covers the mantle feature, but I think just including the other features in your current test should help to overcome this. I'm not sure why some of the testers are failing though... thoughts on this @MFraters

MFraters:
apparently it is not fixed on the tester.... Let's look at it again tomorrow.

MFraters:
hmm, I think I accidentally rebased onto an older main, resulting in a large diff for the declarations and a not accurate coverage result. I can fix this, since it was my mistake.

MFraters:
Fixed the declaration files issues. 

The test that you made is not actually sampling the plume, subducting plate and oceanic plate. Can you either add points in those features, so that we know that they are tested, or you can also change to a gwb grid tests, which may be easier to sample all features.

Other than that, I think it is good to merge :)

Wang-yijun:
> Fixed the declaration files issues.
> 
> The test that you made is not actually sampling the plume, subducting plate and oceanic plate. Can you either add points in those features, so that we know that they are tested, or you can also change to a gwb grid tests, which may be easier to sample all features.
> 
> Other than that, I think it is good to merge :)

Thanks Menno! I've fixed the tests for coverage and also a small old typo, so that the typo tester doesn't break.

MFraters (approved):
Looks great, thanks!

PR #921, opened 2026-03-30
Cookbook for Perlin noise

TilmanMay:
Added a cookbook explaining the perlin noise feature, as Menno said, this would be good for documentation purposes.
#916  followup

### DIFF
--- a/cookbooks/3d_cartesian_perlin_noise/doc/README.md
+++ b/cookbooks/3d_cartesian_perlin_noise/doc/README.md
@@ -36,10 +36,10 @@ Perlin noise is a gradient-based noise function that produces smooth, natural-lo
 
 - `min temperature` / `max temperature`: Set the output value range for the noise field.
 - `octaves`: Number of noise layers combined; more octaves add detail and complexity.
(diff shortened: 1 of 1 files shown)


cookbooks/3d_cartesian_perlin_noise/doc/README.md
```diff
@@ -0,0 +1,95 @@
+(part:user_manual:chap:cookbooks:sec:3d_cartesian_perlin_noise)=
+3D Cartesian Perlin Noise
+========================
+
+This cookbook demonstrates how to set up a 3D Cartesian model in World Builder using Perlin noise to generate spatially varying temperature and composition fields.
+
+
+Normal temperature field without added disturbances
```
danieldouglas92: I imagine this is supposed to be a figure caption, but in the built docs it does not render that way. To have this be a figure caption, you have to include it between the :::   :::, so it would look like this:

```
:::{figure} normal_field.png
:name: perlin_noise_normal_field
:alt: Normal temperature field (without noise)
:align: center

Normal temperature field without added disturbances
:::
```

PR #922, opened 2026-03-30
Added perlin noise feature for topography

TilmanMay:
Add the option to include perlin noise for the topography


danieldouglas92 (commented):
Thanks for another cool extension of the Perlin Noise functionality! Once again only small comments about documentation and consistency between different features, but once these are addressed I think this is good to go.

In a follow up PR (after your cookbook gets merged), I think it would be great to add another section that documents how you would use this feature in that same cookbook as well.

PR #926, opened 2026-03-30
[Obsolete] A new composition parser for all features

alanjyu:
WARNING: "Parameter::composition_properties" (struct) has been renamed to "composition_property" (storing index, name, density, etc. for one composition), while the global composition_properties (vector) indicates the user defined table. These two were named the same in the previous push request.

Users now can assign a composition field using a name (string) that they define in the composition_properties table. The name will be parsed and linked to the corresponding composition index (and default properties). The name output function (what you see in ParaView) has not been updated yet.

### DIFF
--- a/doc/world_builder_declarations.schema.json
+++ b/doc/world_builder_declarations.schema.json
@@ -853,11 +853,15 @@
                                                     ]
                                                 },
                                                 "operation": {
-                                                    "default value": "replace",
+                                                    "default value": "add",
                                                     "type": "string",
-                                                    "description": "Whether the value should replace any value previously defined at this location (replace), add the value to the previously define value (add) or subtract the value to the previously define value (subtract).",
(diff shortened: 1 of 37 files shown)


doc/world_builder_declarations.tex
danieldouglas92: You'll have to delete this file (hopefully won't be an issue ever again after #927 gets merged 🤞

danieldouglas92 (commented):
Hey @alanjyu thanks for this! If I'm understanding correctly, a lot of these changes here are just renaming what you added in PR #909, and then the new get_vector() function you added in parameters.cc is the new feature here? I didn't get a chance to look at #909, but I think being able to set compositions in the world builder file with a string will be a very nice quality of life improvement. I've forgotten so many times what index number N in some of my large world builder files is actually supposed to correspond to.

Anyway, if what I said above is correct and it's not too much of a hassle, I would just open up a very quick PR that only renames `Parameters::composition_properties` to `Parameters::composition_property`.

I think the new parser function looks good, I only had a clarification comment about the modification you made to the existing `get_vector` parser.

alanjyu:
Changes are re-applied in a new pull request #934.

PR #934, opened 2026-03-31
Feature-level composition parser

alanjyu:
Following #909, #926 (obsolete), and #932, users now can assign compositions using an array of names (string) or indices (unsigned int) that they define in the composition_properties map. Name-based entries are parsed and resolved to their corresponding composition indices (and associated default properties)

Although the schema and tests are currently valid, I have not yet identified edge cases where parsing may fail for random composition profiles. In addition, the output naming function has not been updated yet, so visualization still shows Composition_[n] labels.


danieldouglas92:
Thanks for breaking up #926, this makes the current new functionality you're adding here much easier to digest

include/world_builder/parameters.h
```diff
@@ -181,13 +203,6 @@ namespace WorldBuilder
       bool
       check_entry(const std::string &name) const;
 
-      struct composition_property
-      {
```
alanjyu, line 185: Moved up since the struct must be declared before get_vector()

tests/data/composition_properties_map.wb
```diff
@@ -22,12 +22,28 @@
   ],
   "features":[
     {
-      "model":"continental plate",
+      "model":"oceanic plate",
       "name":"TestPlate",
       "max depth":300e3,
-      "coordinates":[[0,0],[100e3,0],[100e3,100e3],[0,100e3]],
+      "coordinates":[[0,0],[100e3,0],[100e3,50e3],[0,50e3]],
```
alanjyu, line 28: Add a test case for different plate models

MFraters:
Can you split out the fix typos into a seperate pull request (make a new branch from main, and you can use `git cherry-pick commit hash`). That make reviewing this pull request easier :)

alanjyu:
I squashed my own typos into [Extend the parser to all features](https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/934/commits/a12a8c6d39fb49ab7584380eaac80db042164361). The typo in the installation_methods.md is left as is. 

Maybe I should drop the Git GUI completely :).

alanjyu:
On my machine with Windows Subsystem for Linux it compiles just fine, can't figure out why it fails some tests here. I am afraid to keep pushing commits just to test.

MFraters:
> On my machine with Windows Subsystem for Linux it compiles just fine, can't figure out why it fails some tests here. I am afraid to keep pushing commits just to test.

Don't be afraid, push as much as you need to find and fix the issue. The testers automatically stop and restart if you push new commits.

I think the problem is that you need to include world.h (or where ever `composition_properties` is defined) into uniform.cc, but I haven't tested it.

MFraters:
You might be able to reproduce the issue if you turn off the unity build. You can do that with `ccmake .` in your build folder. You will get a tui (terminal ui) and you can set `WB_UNITY_BUILD` to OFF by navigating to it and pressing enter. Then use `c` to configure and `g` to generate. After that try recompiling again.

lhy11009:
Hi @alanjyu, I wonder if you still have in your plan to wrap this up. This is already in very shape to me now. I asked this because I have in my other PR https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/886 everything ready except the utilities I need from this PR.

I discussed with @MFraters , we could take over the really small task left if you don't have plan to work on this soon.

PR #943, opened 2026-07-21
Optimize allocations (get them out of loops) with the help of heaptrack.

MFraters:
Based on advice from @bangerth, I played around with heaptrack. I started with the largest allocations entry being 1619842 allocations, and I got it down to ~~43936 (in Rapidjson)~~ update: 795947 allocations.

### DIFF
--- a/source/gwb-grid/main.cc
+++ b/source/gwb-grid/main.cc
@@ -1915,6 +1915,7 @@ int main(int argc, char **argv)
           pool.parallel_for(0, n_p, [&] (size_t i)
           {
             thread_local static  std::vector<double> output(world->properties_output_size(properties));
+            output.clear();
 
             const std::array<double,2> coords = {{grid_x[i], grid_z[i]}};
             world->properties(coords, grid_depth_wrt_surface[i],properties,output);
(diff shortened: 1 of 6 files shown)


source/gwb-grid/main.cc
```diff
@@ -633,6 +634,7 @@ int main(int argc, char **argv)
       /**
        * Begin making the grid
        */
+      std::vector<std::array<unsigned int,3>> topo_input = {{{6,0,0}}};
```
bangerth: You can make that `const` as well.

source/world_builder/world.cc
```diff
@@ -515,9 +514,9 @@ namespace WorldBuilder
     const Objects::NaturalCoordinate natural_coordinate = Objects::NaturalCoordinate(point,*(this->parameters.coordinate_system));
 
     // create output vector
-    std::vector<double> output;
-    std::vector<size_t> entry_in_output;
-    std::vector<std::array<unsigned int,3>> properties_local;
+    std::vector<double> output(World::properties_output_size(properties));
+    std::vector<size_t> entry_in_output(properties.size());
+    std::vector<std::array<unsigned int,3>> properties_local(properties.size());
```
bangerth: This is not the same. You used to create these objects empty, but now you give them a size. Did you mean to create them empty and then call `.reserve()` instead?

PR #945, opened 2026-07-26
Add a maximum topography for external usage

Minerallo:
Hi everyone,

this PR adds `World::maximum_topography()`. It returns an upper bound for the
topography configured in World Builder. This allows applications such as
ASPECT to determine the required geometry size before creating the mesh.

I added tests for a model with uniform topography and for a model without
topography. The complete unit test suite passes.

Cheers,  
Michael

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Added Perlin noise models for composition and temperature across features, and gwb-dat coverage tests for Cartesian Perlin noise inputs. \[Tilman May; 2026-03-29; [#906](https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/906)\]
 
+- Added a `maximum_topography()` API that returns a guaranteed upper bound for all configured topography models. \[Michael Pons; 2026-07-26\]
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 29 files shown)


include/world_builder/features/interface.h
```diff
@@ -88,6 +88,13 @@ namespace WorldBuilder
                         const std::vector<size_t> &entry_in_output,
                         std::vector<double> &output) const = 0;
 
+        /**
+         * Return a guaranteed upper bound for the topography produced by this
+         * feature. Features without topography return zero.
+         */
+        virtual
+        double maximum_topography() const;
```
MFraters: I think every  plugin should implement this, because there is no sensible default implementation. So could you make it `const = 0;`?

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Added Perlin noise models for composition and temperature across features, and gwb-dat coverage tests for Cartesian Perlin noise inputs. \[Tilman May; 2026-03-29; [#906](https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/906)\]
 
+- Added a `maximum_topography()` API that returns a guaranteed upper bound for all configured topography models. \[Michael Pons; 2026-07-26\]
```
MFraters: Can you add this pull request as a reference at the end?

source/world_builder/features/interface.cc
```diff
@@ -63,6 +63,12 @@ namespace WorldBuilder
     Interface::~Interface ()
       = default;
 
+    double
+    Interface::maximum_topography() const
+    {
+      return 0.0;
+    }
+
```
MFraters: I don't think the interface should define a default version. I think every plugin should define a correct value.

MFraters (commented):
Looks generally good, but I have a few comments. I would like t ask you to add tests for every plugin and also add a test to show that you actually picked the correct max topography. 

Would you also be willing to add a minimum topography function, while you are add it?

include/world_builder/features/interface.h
Minerallo: Done. now its double maximum_topography() const = 0;

